PR #52, opened 2023-10-30
[ADD] Support `CrossEntropyLoss` in KFAC

### LINKED ISSUE #44: Support `nn.CrossEntropyLoss` in KFAC

f-dangel:
Closes #44.

As a test, I am considering a data set consisting of a single sample. In that case, the Fisher-blocks are exactly Kronecker-factored, because there is no sum over multiple data points. In the limit of infinite MC-samples, KFAC thus approaches the block-diagonal Fisher.

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -16,9 +16,10 @@ from __future__ import annotations
 from math import sqrt
 from typing import Dict, Iterable, List, Tuple, Union
 
+from einops import rearrange
 from numpy import ndarray
 from torch import Generator, Tensor, einsum, randn
-from torch.nn import Linear, Module, MSELoss, Parameter
+from torch.nn import CrossEntropyLoss, Linear, Module, MSELoss, Parameter
 from torch.utils.hooks import RemovableHandle
 
 from curvlinops._base import _LinearOperator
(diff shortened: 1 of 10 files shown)


curvlinops/kfac.py
```diff
@@ -279,6 +280,23 @@ def draw_label(self, output: Tensor) -> Tensor:
...
+        elif isinstance(self._loss_func, CrossEntropyLoss):
+            # TODO For output.ndim > 2, the scale of the 'would-be' gradient resulting
+            # from these labels might be off
+            if output.ndim != 2:
+                raise NotImplementedError(
+                    "Only 2D output is supported for CrossEntropyLoss for now."
+                )
+            probs = output.softmax(dim=1)
+            # each row contains a vector describing a categorical
+            probs_as_mat = rearrange(probs, "n c ... -> (n ...) c")
```
runame, line 293: The `softmax` and `rearrange` operation could just be executed once before the for-loop over `range(self._mc_samples)` in a `_preprocess_model_output(output)` function or so.

runame (approved):
Besides the potential minor performance improvement in my comment, LGTM.

curvlinops/kfac.py
f-dangel: Hm. Not sure if this will be relevant in practise, as I believe `mc_samples` is almost always `1`, at least for the use cases I know.

PR #54, opened 2023-10-31
[ADD] Support for empirical Fisher in KFAC

### LINKED ISSUE #46: Support using the empirical distribution for gradient covariances in KFAC

runame:
Resolves #46.

I have already added a `'type-2'` option for the `fisher_type` argument. Not sure if I like the name since this terminology is not very common, but using `'exact'` instead lead to confusion before.

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -117,13 +117,14 @@ class KFACLinearOperator(_LinearOperator):
                 from the model's predictive distribution. Defaults to ``2147483647``.
             fisher_type: The type of Fisher/GGN to approximate. If ``'type-2'``, the
                 expectation over the model outputs is computed exactly by
-                computing the backwward pass for each output dimension. This is
+                computing the backward pass for each output dimension. This is
                 sometimes also called type-2 Fisher. If ``'mc'``, the expectation
                 is approximated by sampling ``mc_samples`` labels from the model's
                 predictive distribution. If ``'empirical'``, the empirical gradients
                 are used which corresponds to the uncentered gradient covariance, or
                 the empirical Fisher. Defaults to ``'mc'``.
             mc_samples: The number of Monte-Carlo samples to use per data point.
+                Will be ignored when ``fisher_type`` is not ``'mc'``.
                 Defaults to ``1``.
 
         Raises:
@@ -226,7 +227,6 @@ class KFACLinearOperator(_LinearOperator):
             NotImplementedError: If ``fisher_type == 'type-2'``.
             ValueError: If ``fisher_type`` is not ``'type-2'``, ``'mc'``, or
                 ``'empirical'``.
-
         """
         # install forward and backward hooks
         hook_handles: List[RemovableHandle] = []
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -80,6 +80,7 @@ class KFACLinearOperator(_LinearOperator):
         check_deterministic: bool = True,
         shape: Union[Tuple[int, int], None] = None,
         seed: int = 2147483647,
+        fisher_type: str = "mc",
         mc_samples: int = 1,
     ):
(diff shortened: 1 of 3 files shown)


curvlinops/kfac.py
```diff
@@ -114,6 +115,14 @@ def __init__(
                 from the parameters. Defaults to ``None``.
             seed: The seed for the random number generator used to draw labels
                 from the model's predictive distribution. Defaults to ``2147483647``.
+            fisher_type: The type of Fisher/GGN to approximate. If ``'type-2'``, the
+                expectation over the model outputs is computed exactly by
+                computing the backwward pass for each output dimension. This is
```
f-dangel, line 120: Typo: backwward

curvlinops/kfac.py
```diff
@@ -114,6 +115,14 @@ def __init__(
...
+                sometimes also called type-2 Fisher. If ``'mc'``, the expectation
```
f-dangel, line 121: This description is correct for `MSELoss`, but not for `CrossEntropyLoss`.
- For `MSELoss`, the expected Hessian of the loss w.r.t. the model's output under the model's distribution is `\nabla_f^2 \ell \propto I`. Then, we take each column `c_i \propto e_i`  of `I`, which is a one-hot vector, and compute  `\nabla (f^\top c_i)` for all values of `i`. Hence, this is like calling `.backward` on `f[0], f[1], ...`.
- For `CrossEntropyLoss`, the expected Hessian is `\nabla_f^2 \ell = diag(p) - p p^T` where `p = softmax(f)`. In this case, taking a column does not yield a one-hot vector. Hence, this is not calling `.backward` on `f[0], f[1], ...`. It is calling `backward` on `(\nabla_f^2 \ell)[:, i]` for `i=0, 1, ...`.

curvlinops/kfac.py
```diff
@@ -114,6 +115,14 @@ def __init__(
...
+                is approximated by sampling ``mc_samples`` labels from the model's
+                predictive distribution. If ``'empirical'``, the empirical gradients
+                are used which corresponds to the uncentered gradient covariance, or
+                the empirical Fisher. Defaults to ``'mc'``.
```
f-dangel, line 125: Could mention that `mc_samples` is ignored for empirical and type-2.

curvlinops/kfac.py
```diff
@@ -210,7 +220,14 @@ def _adjoint(self) -> KFACLinearOperator:
...
 
     def _compute_kfac(self):
-        """Compute and cache KFAC's Kronecker factors for future ``matvec``s."""
+        """Compute and cache KFAC's Kronecker factors for future ``matvec``s.
+
+        Raises:
+            NotImplementedError: If ``fisher_type == 'type-2'``.
+            ValueError: If ``fisher_type`` is not ``'type-2'``, ``'mc'``, or
+                ``'empirical'``.
+
```
f-dangel, line 229: Remove blank line.

f-dangel (approved):
Looks good to me modulo the comments on documentation. Remind me whether there is already an issue to implement Fisher type-2.

curvlinops/kfac.py
runame: Added this in the description of `mc_samples` instead of `fisher_type`.

curvlinops/kfac.py
runame: I see now that the wording is ambiguous. I meant to simply describe the overhead by stating the number of backward passes that is required for the selected `fisher_type`. In both cases, for the `MSELoss` and `CrossEntropyLoss`, this is equal to the output dimension/the number of classes for `'type-2'`. I did not want to add any details on the maths or the specific implementation.

Is this clearer?

> If ``'type-2'``, the exact Hessian of the loss w.r.t. the model outputs is used. This requires as many backward passes as the output dimension, i.e. the number of classes for classification.

f-dangel: Yes, that's great!

PR #56, opened 2023-11-07
[ADD] Support for KFAC with type-2 Fisher

### LINKED ISSUE #49: Support type-2 Fisher in KFAC
As opposed to using MC samples from the model's predictive distribution.

runame:
Resolves #49.

Note: I can't immediately see anymore why the implementation for `CrossEntropyLoss` works like this, will have to think about it, but should be correct.

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -261,7 +261,7 @@ class KFACLinearOperator(_LinearOperator):
             handle.remove()
 
     def _compute_loss_and_backward(self, output: Tensor, y: Tensor):
-        """Compute the loss and the backward pass(es) required for KFAC.
+        r"""Compute the loss and the backward pass(es) required for KFAC.
 
         Args:
             output: The model's prediction
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -19,12 +19,8 @@ from typing import Dict, Iterable, List, Set, Tuple, Union
 
 from einops import rearrange
 from numpy import ndarray
-from torch import Generator, Tensor, cat, einsum
-from torch import mean as torch_mean
-from torch import no_grad, randn, stack
-from torch import sum as torch_sum
+from torch import Generator, Tensor, cat, einsum, randn, stack
 from torch.nn import CrossEntropyLoss, Linear, Module, MSELoss, Parameter
-from torch.nn.functional import log_softmax, softmax
 from torch.utils.hooks import RemovableHandle
 
 from curvlinops._base import _LinearOperator
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -262,12 +262,12 @@ class KFACLinearOperator(_LinearOperator):
 
     def _compute_loss_and_backward(self, output: Tensor, y: Tensor):
         """Compute the loss and the backward pass(es) required for KFAC.
-        
+
         Args:
             output: The model's prediction
                 :math:`\{f_\mathbf{\theta}(\mathbf{x}_n)\}_{n=1}^N`.
             y: The labels :math:`\{\mathbf{y}_n\}_{n=1}^N`.
-        
+
         Raises:
             NotImplementedError: If ``fisher_type == 'type-2'`` and
                 ``isinstance(self._loss_func, CrossEntropyLoss)``.
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -21,13 +21,14 @@ from einops import rearrange
 from numpy import ndarray
 from torch import Generator, Tensor, cat, einsum
 from torch import mean as torch_mean
-from torch import no_grad, randn
+from torch import no_grad, randn, stack
 from torch import sum as torch_sum
 from torch.nn import CrossEntropyLoss, Linear, Module, MSELoss, Parameter
 from torch.nn.functional import log_softmax, softmax
 from torch.utils.hooks import RemovableHandle
 
 from curvlinops._base import _LinearOperator
+from curvlinops.kfac_utils import hessian_matrix_sqrt
 
 
 class KFACLinearOperator(_LinearOperator):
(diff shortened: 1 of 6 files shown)


curvlinops/kfac.py
```diff
@@ -265,31 +268,63 @@ def _compute_kfac(self):
...
+            elif isinstance(self._loss_func, CrossEntropyLoss):
+                flat_logits = output.flatten(end_dim=-2)
+                log_probs = log_softmax(flat_logits, dim=-1)
+                with no_grad():
+                    sqrt_probs = softmax(flat_logits, dim=-1).sqrt()
+                num_classes = log_probs.shape[1]
+                for c in range(num_classes):
+                    # Mean or sum reduction over all loss terms.
+                    loss_c = reduction_fn(-log_probs[:, c] * sqrt_probs[:, c])
+                    loss_c.backward(retain_graph=c < num_classes - 1)
```
f-dangel, line 313: I will try to refactor this in terms of the loss Hessian's matrix square root.

f-dangel: Done, please take a look.

f-dangel:
I decided to start a section 'Internals' in the documentation which explains the tricky parts in math and also contains links to derivations. This might be useful for others who are interested in extending this library, or learning more about the details of each curvature approximation. 

I have worked on this topic for quite a bit, so it would be great to get feedback if it is comprehensible.

You can take a look [here](https://curvlinops--56.org.readthedocs.build/en/56/internals.html#).

curvlinops/kfac_utils.py
```diff
@@ -0,0 +1,92 @@
+"""Utility functions related to KFAC."""
+
+from math import sqrt
+from typing import Union
+
+from torch import Tensor, diag, einsum, eye
+from torch.nn import CrossEntropyLoss, MSELoss
+
+
+def hessian_matrix_sqrt(
```
runame, line 10: Could be a bit more explicitly named as `loss_hessian_matrix_sqrt`.

curvlinops/kfac_utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+        .. math::
+            \ell(\mathbf{f}) &= c \sum_{i=1}^C (f_i - y_i)^2
+            \\
+            \nabla^2_{\mathbf{f}} \ell(\mathbf{f}, \mathbf{y}) &= 2 c \mathbf{I}_C
+            \\
+            \mathbf{S} &= \sqrt{2 c} \mathbf{I}_C
+
+    Note:
+        For :class:`torch.nn.CrossEntropyLoss` (with :math:`c = 1` irrespective of the
```
runame, line 48: Why introduce `c` here when it is always set to `1` anyway?

curvlinops/kfac.py
```diff
@@ -266,31 +266,70 @@ def _compute_kfac(self):
...
+        Raises:
+            ValueError: If ``fisher_type`` is not ``'type-2'``, ``'mc'``, or
+                ``'empirical'``.
+            NotImplementedError: If ``fisher_type`` is ``'type-1'`` and the
+                output is not 2d.
+        """
+        if self._fisher_type == "type-2":
+            if output.ndim != 2:
+                raise NotImplementedError(
+                    "Type-2 Fisher not implemented for non-2d output."
```
runame, line 293: Why not? I think we know how to reshape to 2d to get the desired quantity.

curvlinops/kfac.py
```diff
@@ -266,31 +266,70 @@ def _compute_kfac(self):
...
+                )
+            # Compute per-sample Hessian square root, then concatenate over samples.
+            # Result has shape `(batch_size, num_classes, num_classes)`
+            hessian_sqrts = stack(
```
runame, line 297: Maybe it would be cleaner to handle the potential batching inside of `hessian_matrix_sqrt`.

curvlinops/kfac_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    output = output_one_datum.squeeze(0)
+    num_classes = output.numel()
+
+    if isinstance(loss_func, MSELoss):
+        c = {"sum": 1.0, "mean": 1.0 / num_classes}[loss_func.reduction]
+        return eye(num_classes, device=output.device, dtype=output.dtype).mul_(
+            sqrt(2 * c)
+        )
+    elif isinstance(loss_func, CrossEntropyLoss):
+        c = 1.0
```
runame, line 87: When is this going to change?

curvlinops/kfac_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        ValueError: If the batch size is not one, or the output is not 2d.
+        NotImplementedError: If the loss function is not supported.
+    """
+    if output_one_datum.ndim != 2 or output_one_datum.shape[0] != 1:
+        raise ValueError(
+            f"Expected 'output_one_datum' to be 2d with shape [1, C], got "
+            f"{output_one_datum.shape}"
+        )
+    output = output_one_datum.squeeze(0)
+    num_classes = output.numel()
```
runame, line 79: `num_classes` is a bit misleading for regression.

curvlinops/kfac_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    if isinstance(loss_func, MSELoss):
+        c = {"sum": 1.0, "mean": 1.0 / num_classes}[loss_func.reduction]
+        return eye(num_classes, device=output.device, dtype=output.dtype).mul_(
+            sqrt(2 * c)
+        )
+    elif isinstance(loss_func, CrossEntropyLoss):
+        c = 1.0
+        p = output_one_datum.softmax(dim=1).squeeze()
+        p_sqrt = p.sqrt()
+        return (diag(p_sqrt) - einsum("i,j->ij", p, p_sqrt)).mul_(sqrt(c))
```
runame, line 90: Here, we explicitly build a `(num_classes, num_classes)` matrix (same for regression). However, we backpropagate this matrix column wise -- is it not more efficient to also just build it column-wise before backpropagating? Especially in the regression case this seems obvious, since it will only be unit vectors times a scalar, no need to ever build the full identity matrix.

runame (commented):
I really like the refactor in terms of backpropagating the square root loss hessian and the idea of having an internals section in the docs! I left a few comments/questions.

curvlinops/kfac_utils.py
f-dangel: Agree. Done.

curvlinops/kfac.py
f-dangel: Would that be tested somewhere? This is is the only concern that lead me to restricting this for now.

curvlinops/kfac.py
f-dangel: I thought about that too.

In my opinion, I prefer to separate the unbatched behavior from the batched behavior for clarity, although this will lead to a for loop internally.

This also helped me to understand that there are **two** reductions. One is in the case of batches, the other in the case of vector/sequence-valued outputs.

This is my personal optinion. In BackPACK, everything is handled together for performance, but I'm actually not sure if those benefits outweigh the more complex implementation (to explain this to someone else, all the quantities would get an additional batch dimension).

I'm okay changing this if it becomes a crucial performance bottleneck.

curvlinops/kfac_utils.py
f-dangel: Would you prefer `C` (as used in the math) over `num_classes`?

curvlinops/kfac_utils.py
f-dangel: Yes, you are completely right about this.

Again, this is a complexity-vs-performance trade-off. I believe `num_classes` will be `< 10_000` because otherwise we would have to perform `>= 10_000` backpropagations. 

I'm okay with paying the price of having to store `batch_size` matrices of size `(num_classes, num_classes)`; for now.

curvlinops/kfac_utils.py
runame: Yeah, I think I would prefer `C` or sth like `output_dim`.

curvlinops/kfac.py
runame: Have to look into how you implemented the `GGNLinearOperator`. Does that also only support 2d outputs? Otherwise you're right and we should add more tests + >2d outputs in a future PR.

runame: Have created an issue to track this: #60.

curvlinops/kfac.py
runame: Ok, I'm fine leaving this like it is for now.

curvlinops/kfac_utils.py
runame: Ok, let's leave this for now.

curvlinops/kfac_utils.py
f-dangel: Agree that we could do that in the future. However, right now this would violate the convention to separate effects from `batch_size > 1`.

runame: Although you can interpret all additional output dimensions as an extension of the batch size, since the loss just treats them as additional i.i.d. samples.

PR #57, opened 2023-11-07
[ADD] Option to treat weight and bias of a layer jointly in KFAC

### LINKED ISSUE #53: Support treating weights and biases of a layer jointly in KFAC

f-dangel:
Eventually, this will resolve #53.

At the moment, I am struggling to get the tests running and would benefit from a second pair of eyes to look through the code. At the moment, the tests seem to execute for both `separate_weight_and_bias = [False, True]`, but the matrix produced by KFAC is wrong for `True`.

I believe this is related to the order of parameters. But I think I am actually arranging them in the correct order 😂, so maybe I'm barking up the wrong tree. In any case, it would be great if you could skim the code and look for anything obvious I missed.

### DIFF
--- a/test/test_kfac.py
+++ b/test/test_kfac.py
@@ -57,7 +57,14 @@ def test_kfac(
         separate_weight_and_bias=separate_weight_and_bias,
     )
 
-    kfac = KFACLinearOperator(model, loss_func, params, data, mc_samples=2_000)
+    kfac = KFACLinearOperator(
+        model,
+        loss_func,
+        params,
+        data,
+        mc_samples=2_000,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
     kfac_mat = kfac @ eye(kfac.shape[1])
 
     atol = {"sum": 5e-1, "mean": 5e-3}[loss_func.reduction]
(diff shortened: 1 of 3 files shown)


f-dangel:
**Update:** I found the problem and fixed everything. Ready for review.

test/test_kfac.py
```diff
@@ -9,11 +10,11 @@
 from torch.nn import Module, MSELoss, Parameter
 
 from curvlinops.examples.utils import report_nonclose
-from curvlinops.ggn import GGNLinearOperator
 from curvlinops.gradient_moments import EFLinearOperator
 from curvlinops.kfac import KFACLinearOperator
 
 
+@mark.parametrize("separate_weight_and_bias", [True, False], ids=["separate", "joint"])
```
runame, line 17: The `ids` could a bit more verbose here, e.g. `ids=["separate_bias", "joint_bias"]`.

runame (approved):
Sorry was asleep :D But LGTM!

test/test_kfac.py
f-dangel: Incorporated.

PR #59, opened 2023-11-08
[ADD] Support KFAC if the network has inplace activations

### LINKED ISSUE #58: Bug: Networks with in-place operations not supported in KFAC
This is similar to https://github.com/f-dangel/singd/issues/56 and caused by the implementation of computing the gradient covariance via `full_backward_hook`s. The fix is to replace the module backward hook with a tensor backward hook installed on the output of a layer.

f-dangel:
Resolves #58.

### DIFF
--- a/test/test_kfac.py
+++ b/test/test_kfac.py
@@ -1,12 +1,14 @@
 """Contains tests for ``curvlinops.kfac``."""
 
+from test.cases import DEVICES, DEVICES_IDS
+from test.utils import regression_targets
 from typing import Iterable, List, Tuple
 
 from numpy import eye
 from pytest import mark
 from scipy.linalg import block_diag
-from torch import Tensor, randperm
-from torch.nn import Module, MSELoss, Parameter
+from torch import Tensor, device, manual_seed, rand, randperm
+from torch.nn import Linear, Module, MSELoss, Parameter, ReLU, Sequential
 
 from curvlinops.examples.utils import report_nonclose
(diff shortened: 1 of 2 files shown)


test/test_kfac.py
```diff
@@ -103,3 +105,49 @@ def test_kfac_ef_one_datum(
...
+
+    kfac = KFACLinearOperator(model, loss_func, params, data, mc_samples=2_000)
+    kfac_mat = kfac @ eye(kfac.shape[1])
+
+    atol = {"sum": 5e-1, "mean": 2e-3}[loss_func.reduction]
+    rtol = {"sum": 2e-2, "mean": 2e-2}[loss_func.reduction]
+
+    report_nonclose(ggn, kfac_mat, rtol=rtol, atol=atol)
+
+    # 2) Compare GGN (inplace=True) and GGN (inplace=False)
```
runame, line 142: Shouldn't we here compare "GGN (inplace=True) against KFAC (inplace=False)" or equivalently "KFAC (inplace=True) against KFAC (inplace=False)".

runame (changes requested):
Can't we also apply the same refactor as in SINGD, where we just use a single forward hook that computes the input covariance and installs the tensor hook on the module output?

f-dangel:
One point against merging these hooks is that we may want to re-cycle the hook for computing the activation variance (e.g. FOOF).

runame:
Ok let's stick to separate hooks for now then.

runame (approved):

Issue #50, opened 2023-10-30
Support KFAC-expand and KFAC-reduce for linear weight-sharing layers.


runame:
Specifically, for `nn.Linear` and `nn.Conv2d` (after #45) layers.

PR #62, opened 2024-01-09
Support KFAC for >2d model outputs

### LINKED ISSUE #60: Support >2d model outputs for KFAC with type-2 Fisher
See [this thread](https://github.com/f-dangel/curvlinops/pull/56#discussion_r1388083449).

runame:
Resolves #60.

This implements K-FAC-expand for linear layers with >2d inputs. Will address #50 in a follow-up PR. Also, I already added the `loss_average` keyword for the correct scaling of the gradient covariance.

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -71,6 +71,11 @@ class KFACLinearOperator(_LinearOperator):
 
     _SUPPORTED_LOSSES = (MSELoss, CrossEntropyLoss)
     _SUPPORTED_MODULES = (Linear,)
+    _SUPPORTED_LOSS_AVERAGE: Tuple[Union[None, str], ...] = (
+        None,
+        "batch",
+        "batch+sequence",
+    )
 
     def __init__(
         self,
@@ -84,6 +89,7 @@ class KFACLinearOperator(_LinearOperator):
         seed: int = 2147483647,
         fisher_type: str = "mc",
         mc_samples: int = 1,
+        loss_average: Union[None, str] = "batch",
         separate_weight_and_bias: bool = True,
     ):
         """Kronecker-factored approximate curvature (KFAC) proxy of the Fisher/GGN.
@@ -128,6 +134,15 @@ class KFACLinearOperator(_LinearOperator):
             mc_samples: The number of Monte-Carlo samples to use per data point.
                 Has to be set to ``1`` when ``fisher_type != 'mc'``.
                 Defaults to ``1``.
+            loss_average: Whether the loss function is a mean over per-sample
+                losses and if yes, over which dimensions the mean is taken.
+                If `"batch"`, the loss function is a mean over as many terms as
+                the size of the mini-batch. If `"batch+sequence"`, the loss
+                function is a mean over as many terms as the size of the
+                mini-batch times the sequence length, e.g. in the case of
+                language modeling. If `None`, the loss function is a sum. This
+                argument is used to ensure that the preconditioner is scaled
+                consistently with the loss and the gradient. Default: `"batch"`.
             separate_weight_and_bias: Whether to treat weights and biases separately.
                 Defaults to ``True``.
 
@@ -138,6 +153,16 @@ class KFACLinearOperator(_LinearOperator):
         if not isinstance(loss_func, self._SUPPORTED_LOSSES):
             raise ValueError(
                 f"Invalid loss: {loss_func}. Supported: {self._SUPPORTED_LOSSES}."
+            )
+        if loss_average not in self._SUPPORTED_LOSS_AVERAGE:
+            raise ValueError(
+                f"Invalid loss_average: {loss_average}. "
+                f"Supported: {self._SUPPORTED_LOSS_AVERAGE}."
+            )
+        if loss_average is None and loss_func.reduction != "sum":
+            raise ValueError(
+                f"Invalid loss_average: {loss_average}. "
+                f"Must be 'batch' or 'batch+sequence' if loss_func.reduction != 'sum'."
             )
         if fisher_type != "mc" and mc_samples != 1:
             raise ValueError(
@@ -167,6 +192,7 @@ class KFACLinearOperator(_LinearOperator):
         self._separate_weight_and_bias = separate_weight_and_bias
         self._fisher_type = fisher_type
         self._mc_samples = mc_samples
+        self._loss_average = loss_average
         self._input_covariances: Dict[str, Tensor] = {}
         self._gradient_covariances: Dict[str, Tensor] = {}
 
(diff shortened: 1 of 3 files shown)


curvlinops/kfac.py
```diff
@@ -128,6 +134,15 @@ def __init__(
...
                 Defaults to ``1``.
+            loss_average: Whether the loss function is a mean over per-sample
+                losses and if yes, over which dimensions the mean is taken.
+                If `"batch"`, the loss function is a mean over as many terms as
+                the size of the mini-batch. If `"batch+sequence"`, the loss
+                function is a mean over as many terms as the size of the
+                mini-batch times the sequence length, e.g. in the case of
+                language modeling. If `None`, the loss function is a sum. This
+                argument is used to ensure that the preconditioner is scaled
+                consistently with the loss and the gradient. Default: `"batch"`.
```
f-dangel, line 145: Small nit because `curvlinops` uses sphinx to render docstrings. You'll have to replace all single backticks with double backticks.

curvlinops/kfac.py
```diff
@@ -365,20 +394,12 @@ def draw_label(self, output: Tensor) -> Tensor:
             return output.clone().detach() + perturbation
 
         elif isinstance(self._loss_func, CrossEntropyLoss):
-            # TODO For output.ndim > 2, the scale of the 'would-be' gradient resulting
-            # from these labels might be off
-            if output.ndim != 2:
```
f-dangel, line 370: Do you think it makes sense to add a safeguard at the beginning of this function that `assert`s `output.ndim==2`?

I like the idea of collecting all axes that are treated like batch axes in a single dimension. Just want to make sure that we remember we're using this idea whenever we use `draw_label` in the future (potentially for other curvature approximations).

f-dangel: Also, making this assumption within `draw_label` simplifies the code for `MSELoss`. If we decide to make this assumption, can you refactor the code handling `MSELoss`, too?

curvlinops/kfac.py
```diff
@@ -365,20 +394,12 @@ def draw_label(self, output: Tensor) -> Tensor:
...
-                raise NotImplementedError(
-                    "Only 2D output is supported for CrossEntropyLoss for now."
-                )
-            probs = output.softmax(dim=1)
             # each row contains a vector describing a categorical
-            probs_as_mat = rearrange(probs, "n c ... -> (n ...) c")
```
f-dangel, line 376: Maybe remove the comment above as it is outdated?

test/utils.py
```diff
@@ -24,13 +25,28 @@ def get_available_devices():
     return devices
 
 
-def classification_targets(size, num_classes):
-    """Create random targets for classes 0, ..., `num_classes - 1`."""
+def classification_targets(size: Tuple[int], num_classes: int) -> Tensor:
+    """Create random targets for classes 0, ..., `num_classes - 1`.
```
f-dangel, line 29: Nit: Double backticks.

test/utils.py
```diff
@@ -91,3 +107,27 @@ def ggn_block_diagonal(
...
+    def forward(self, x: Tensor) -> Tensor:
+        """Rearrange the input tensor.
+
+        Args:
+            x: The input tensor.
+
+        Returns:
+            The rearranged tensor.
+        """
+        return rearrange(x, self.pattern)
```
f-dangel, line 133: I think `einops` has exactly this layer (https://einops.rocks/#layers). Should be `einops.layers.torch.Rearrange`. You can probably use that.

f-dangel (commented):
Still need to take a look at the test files but feel free to start discussing the points.

test/test_kfac.py
```diff
@@ -200,3 +206,71 @@ def test_kfac_inplace_activations(dev: device):
...
+        dev: The device to run the test on.
+    """
+    manual_seed(0)
+    # set up loss function, data, and model
+    loss_func = loss(reduction=reduction).to(dev)
+    if isinstance(loss_func, MSELoss):
+        data = [
+            (rand(2, 4, 5), regression_targets((2, 4, 3))),
+            (rand(4, 7, 5, 5), regression_targets((4, 7, 5, 3))),
+        ]
```
f-dangel, line 236: There is a problem with feeding such data.
`curvlinops` assumes that all batches in the data set can be stacked along the batch axis.
If this is not the case, `curvlinops` cannot figure out how to properly accumulate mini-batch losses into the loss over the entire data set, because it would have to know what axes are being reduced (and how).

This convention that "batches from the dataset need to be stackable along axis 0" is not enforced in the code because it assumes that data from a standardized dataset has the same dimension, only the batch size can vary.

To fix this, you would have to split up the test into a test for a 3d data set and a 4d data set. Let me know if this does not make sense, or if you think this assumption is unrealistic and we need to fix this in other parts of the library.

test/test_kfac.py
```diff
@@ -200,3 +206,71 @@ def test_kfac_inplace_activations(dev: device):
...
+        manual_seed(711)
+        model = Sequential(Linear(5, 4), Linear(4, 3)).to(dev)
+    else:
+        data = [
+            (rand(2, 4, 5), classification_targets((2, 4), 3)),
+            (rand(4, 7, 5, 5), classification_targets((4, 7, 5), 3)),
+        ]
```
f-dangel, line 243: See the comment above.

f-dangel (commented):
Also had a look at the tests.

test/test_kfac.py
runame: Ok I see. I'm a bit confused that KFAC-reduce (type-2) matches the `curvlinops` GGN in the test in #63 for different weight-sharing dims per batch. Does this not violate the assumptions?

curvlinops/kfac.py
runame: Not sure in what sense the code for `MSELoss` can be more simplified (besides changing the `std` for the `"mean"` case; see my latest commit).

f-dangel: I meant that, but when I wrote it down I assumed the diff would be bigger 😂. Thanks for fixing this!

test/test_kfac.py
f-dangel: That might be a lucky coincidence and I will try to think through this while reviewing #63. 

I think I've convinced myself that all `x` from a dataset must have the same shape and therefore my concern above is valid.
If this was not the case, we could not draw random data from the data set and stack them into a mini-batch (because stacking requires equal shape up until the axis to be stacked, which is the batch axis).

curvlinops/kfac.py
```diff
@@ -284,14 +310,17 @@ def _compute_loss_and_backward(self, output: Tensor, y: Tensor):
...
         """
+        # if >2d output we convert to an equivalent 2d output
+        if output.ndim > 2:
+            if isinstance(self._loss_func, CrossEntropyLoss):
+                output = rearrange(output, "batch c ... -> (batch ...) c")
+                y = rearrange(y, "batch ... -> (batch ...)")
+            else:
+                output = rearrange(output, "batch ... c -> (batch ...) c")
+                y = rearrange(y, "batch ... c -> (batch ...) c")
+
```
f-dangel, line 322: Sorry I seem to have missed this in the first review, but we could simply remove the `if` statement checking `output.ndim`. The `rearrange`s won't do anything if `output.ndim==2`.

f-dangel (approved):
Please go ahead and squash-merge after addressing or arguing against my remaining comment. Looks great!

PR #64, opened 2024-01-10
Add support for KFAC for `Conv2d` modules

### LINKED ISSUE #45: Support KFC (Grosse 2016) for convolutional layers in KFAC

runame:
Resolves #45.

This review is based on #62 and #63 which should be reviewed first. I took the `extract_patches` and `extract_averaged_patches` from the SINGD repo, maybe we should add a reference.

_Note: Currently one of the tests fails due to numerical issues (it passes locally). -> Fixed by changing the `rtol` for the KFAC for weight sharing layers test from 1e-5 to 1e-4. Not ideal, but should hopefully be fine._

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -21,11 +21,15 @@ from typing import Dict, Iterable, List, Set, Tuple, Union
 from einops import rearrange, reduce
 from numpy import ndarray
 from torch import Generator, Tensor, cat, einsum, randn, stack
-from torch.nn import CrossEntropyLoss, Linear, Module, MSELoss, Parameter
+from torch.nn import Conv2d, CrossEntropyLoss, Linear, Module, MSELoss, Parameter
 from torch.utils.hooks import RemovableHandle
 
 from curvlinops._base import _LinearOperator
-from curvlinops.kfac_utils import loss_hessian_matrix_sqrt
+from curvlinops.kfac_utils import (
+    extract_averaged_patches,
+    extract_patches,
+    loss_hessian_matrix_sqrt,
+)
 
 
 class KFACLinearOperator(_LinearOperator):
@@ -71,7 +75,7 @@ class KFACLinearOperator(_LinearOperator):
     """
 
     _SUPPORTED_LOSSES = (MSELoss, CrossEntropyLoss)
-    _SUPPORTED_MODULES = (Linear,)
+    _SUPPORTED_MODULES = (Linear, Conv2d)
     _SUPPORTED_LOSS_AVERAGE: Tuple[Union[None, str], ...] = (
         None,
         "batch",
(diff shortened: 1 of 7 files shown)


runame:
I think I would prefer making `curvlinops` a requirement of the `singd` repo because I think the implementation here should be our reference KFAC implementation, whereas SINGD simply uses KFAC, i.e. the repo is not focused on KFAC itself. I even thought about removing all KFAC code from the `singd` repo and using `curvlinops` instead, but the issue is that for SINGD we use `a` and `g` and not their outer products, which is the only thing we get access to in the current implementation in this repo. The motivation for this is to have a single KFAC implementation so that all new features or fixes are automatically available for all methods that use KFAC, e.g. SINGD.

f-dangel:
I think your idea is better.

How about the following:
- We duplicate the KFAC utilities for now
- We open an issue on the SINGD repo whose goal is to refactor the code and remove all KFAC utils in favour of `curvlinops`
- Once we have a new `curvlinops` release, we carry out the above issue.

runame:
Fully agreed, I have created an issue [here](https://github.com/f-dangel/singd/issues/69).

curvlinops/kfac.py
```diff
@@ -529,13 +524,26 @@ def _hook_accumulate_input_covariance(self, module: Module, inputs: Tuple[Tensor
...
         if len(inputs) != 1:
             raise ValueError("Modules with multiple inputs are not supported.")
         x = inputs[0].data.detach()
 
+        if isinstance(module, Conv2d):
+            patch_extractor_fn = (
+                extract_patches
+                if self._kfac_approx == "expand"
+                else extract_averaged_patches
+            )
```
f-dangel, line 537: Nit: Could use a dict which is a bit more explicit but less efficient:
```python
patch_extractor_fn = {
     "expand": extract_patches,
     "reduce": extract_averaged_patches,
}[self._kfac_approx]
```

curvlinops/kfac_utils.py
```diff
@@ -87,6 +92,111 @@ def loss_hessian_matrix_sqrt(
...
+        kernel_size: The convolution's kernel size supplied as 2-tuple or integer.
+        stride: The convolution's stride supplied as 2-tuple or integer.
+        padding: The convolution's padding supplied as 2-tuple, integer, or string.
+        dilation: The convolution's dilation supplied as 2-tuple or integer.
+        groups: The number of channel groups.
+
+    Returns:
+        A tensor of shape `[batch_size, O1 * O2, C_in // groups * K1 * K2]` where
+        each column `[b, o1_o2, :]` contains the flattened patch of sample `b` used
+        for output location `(o1, o2)`, averaged over channel groups.
```
f-dangel, line 123: Nit: Double backticks

curvlinops/kfac_utils.py
```diff
@@ -87,6 +92,111 @@ def loss_hessian_matrix_sqrt(
...
+    padding: Union[Tuple[int, int], int, str],
+    dilation: Union[Tuple[int, int], int],
+    groups: int,
+) -> Tensor:
+    """Extract averaged patches from the input of a 2d-convolution.
+
+    The patches are averaged over channel groups and output locations.
+
+    Uses the tensor network formulation of convolution from
+    [Dangel, 2023](https://arxiv.org/abs/2307.02275).
```
f-dangel, line 159: Nit: Use rst syntax.

f-dangel (approved):
A few minor things. Feel free to address and merge without further feedback 👍

PR #69, opened 2024-02-04
[ADD] `KFACInverseLinearOperator`

### LINKED ISSUE #68: Add `KFACInverseLinearOperator` class
To efficiently invert instances of the `KFACLinearOperator`.

runame:
Resolves #68. Implemented by naively inverting all Kronecker factors with the help of a Cholesky decomposition.

One downside of the implementation of the `_matvec` method is that it repeats basically all of the code of the same method from the `KFACLinearOperator`. To avoid code redundancy we could modify `KFACLinearOperator` in-place, but that seems a bit hacky.

### DIFF
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -212,7 +212,7 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
     def __init__(
         self,
         A: KFACLinearOperator,
-        damping: Optional[Tuple[float, float]] = None,
+        damping: Tuple[float, float] = (0., 0.),
         cache: bool = True,
     ):
         """Store the linear operator whose inverse should be represented.
@@ -220,6 +220,7 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
         Args:
             A: ``KFACLinearOperator`` whose inverse is formed.
             damping: Damping values for all input and gradient covariances.
+                Default: ``(0., 0.)``.
             cache: Whether to cache the inverses of the Kronecker factors.
                 Default: ``True``.
 
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -1,10 +1,25 @@
 """Implements linear operator inverses."""
 
-from numpy import column_stack, ndarray
+from numpy import allclose, column_stack, ndarray
 from scipy.sparse.linalg import LinearOperator, cg
 
 
-class CGInverseLinearOperator(LinearOperator):
+class _InverseLinearOperator(LinearOperator):
+    """Base class for (approximate) inverses of linear operators."""
+
+    def _matmat(self, X: ndarray) -> ndarray:
+        """Matrix-matrix multiplication.
+
+        Args:
+            X: Matrix for multiplication.
+
+        Returns:
+            Matrix-multiplication result ``A⁻¹@ X``.
+        """
+        return column_stack([self @ col for col in X.T])
+
+
+class CGInverseLinearOperator(_InverseLinearOperator):
     """Class for inverse linear operators via conjugate gradients."""
 
     def __init__(self, A: LinearOperator):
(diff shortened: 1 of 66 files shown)


test/test_inverse.py
```diff
@@ -125,3 +128,27 @@ def test_NeumannInverseLinearOperator_toy():
...
+    KFAC = KFACLinearOperator(
+        model_func, loss_func, params, data, loss_average=loss_average
+    )
+    KFAC._compute_kfac()
+    # add damping
+    for aaT, ggT in zip(
+        KFAC._input_covariances.values(), KFAC._gradient_covariances.values()
+    ):
+        aaT += delta * torch.eye(aaT.shape[0])
+        ggT += delta * torch.eye(ggT.shape[0])
```
f-dangel, line 148: Nit: Could use `add_(eye(...), alpha=delta)`.

f-dangel: Nit: I think this should crash on GPU because  the `eye` lives on the incorrect device.

test/test_inverse.py
```diff
@@ -125,3 +128,27 @@ def test_NeumannInverseLinearOperator_toy():
...
+
+    inv_KFAC = KFACInverseLinearOperator(KFAC, cache=cache)
+    inv_KFAC_naive = torch.inverse(torch.as_tensor(KFAC @ torch.eye(KFAC.shape[0])))
+
+    x = random.rand(KFAC.shape[1])
+    report_nonclose(inv_KFAC @ x, inv_KFAC_naive @ x, rtol=1e-3)
```
f-dangel, line 154: This might crash on GPU due to device mismatches.

test/test_inverse.py
runame: Question: is the tests in this file even run on multiple devices? I'm still sometimes confused by the use of `case`.

test/test_inverse.py
runame: Both values will always be `numpy` arrays, so there should be no issue here.

curvlinops/inverse.py
```diff
@@ -197,3 +204,102 @@ def _matvec(self, x: ndarray) -> ndarray:
                 )
 
         return self._scale * result
+
+
+class KFACInverseLinearOperator(_InverseLinearOperator):
+    """Class to invert instances of the ``KFACLinearOperator``."""
+
+    def __init__(self, A: KFACLinearOperator, cache: bool = True):
```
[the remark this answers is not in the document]
runame: Yea, I also considered some functionality to add damping. I haven't implemented anything yet because I was not sure if we should only support adding a fixed value, or use a form of adaptive damping (as used by Kazuki in `ASDL`). For now we can follow your suggestion though.

test/test_inverse.py
f-dangel: You're right.

test/test_inverse.py
f-dangel: If your local computer has a GPU, the tests will be run on both CPU and GPU:
- `case` fixture [loads](https://github.com/f-dangel/curvlinops/blob/ae0d25e04ecd606b32b275fd2f864dcb96ef6bda/test/conftest.py#L28-L29) model and loss function to the device

curvlinops/inverse.py
```diff
@@ -197,3 +204,121 @@ def _matvec(self, x: ndarray) -> ndarray:
...
+            name: Name of the layer for which to invert Kronecker factors.
+
+        Returns:
+            Tuple of inverses of the input and gradient covariance Kronecker factors.
+        """
+        if name in self._inverse_input_covariances:
+            aaT_inv = self._inverse_input_covariances.get(name)
+            ggT_inv = self._inverse_gradient_covariances.get(name)
+        else:
+            aaT = self._A._input_covariances.get(name)
```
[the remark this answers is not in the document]
f-dangel: I.e. `aaT` should only be fetched and inverted if `A.model_func.name.weight` is in params

curvlinops/inverse.py
```diff
@@ -197,3 +204,121 @@ def _matvec(self, x: ndarray) -> ndarray:
...
+
+        Args:
+            name: Name of the layer for which to invert Kronecker factors.
+
+        Returns:
+            Tuple of inverses of the input and gradient covariance Kronecker factors.
+        """
+        if name in self._inverse_input_covariances:
+            aaT_inv = self._inverse_input_covariances.get(name)
+            ggT_inv = self._inverse_gradient_covariances.get(name)
```
f-dangel, line 251: Suggestion: Simply `return aaT_inv, ggT_inv` here, then you can un-indent everything below by one tab.

curvlinops/inverse.py
```diff
@@ -197,3 +204,121 @@ def _matvec(self, x: ndarray) -> ndarray:
...
+            damping_ggT = self._damping[1] if self._damping is not None else 0.0
+            ggT_inv = (
+                cholesky_inverse(
+                    cholesky(ggT + damping_ggT * eye(ggT.shape[0], device=ggT.device))
+                )
+                if ggT is not None
+                else None
+            )
+            if self._cache:
+                self._inverse_input_covariances[name] = aaT_inv
```
f-dangel, line 272: Only if layer's weight is in params.

f-dangel (commented):
I need to take a look at the internals of `KFAC`. I believe that `aaT` should not be computed if a layer's `weight` is not considered. Since the current tests are passing, I am confused about this since it suggests `aaT` is computed in such cases, although it shouldn't.

curvlinops/inverse.py
```diff
@@ -197,3 +204,121 @@ def _matvec(self, x: ndarray) -> ndarray:
...
         return self._scale * result
+
+
+class KFACInverseLinearOperator(_InverseLinearOperator):
+    """Class to invert instances of the ``KFACLinearOperator``."""
+
+    def __init__(
+        self,
+        A: KFACLinearOperator,
+        damping: Optional[Tuple[float, float]] = None,
```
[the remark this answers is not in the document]
runame: True, I originally wanted to avoid multiplication by `0` but ended up with it anyway haha

curvlinops/inverse.py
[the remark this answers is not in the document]
runame: It won't fail since `get` returns `None` when the key is not in the dict. Then `aaT_inv` is `None` in the `_matvec` method and it will only be multiplied with the vector if `weight` is in params, see [here](https://github.com/f-dangel/curvlinops/blob/kfac-inv/curvlinops/inverse.py#L316).

curvlinops/inverse.py
runame: See https://github.com/f-dangel/curvlinops/pull/69#discussion_r1480333199.

curvlinops/inverse.py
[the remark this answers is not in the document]
f-dangel: Great, thanks for clarifying this. I was just worried something was wrong inside `KFAC`.

curvlinops/inverse.py
```diff
@@ -197,3 +204,120 @@ def _matvec(self, x: ndarray) -> ndarray:
...
+            ggT_inv = self._inverse_gradient_covariances.get(name)
+            return aaT_inv, ggT_inv
+        aaT = self._A._input_covariances.get(name)
+        ggT = self._A._gradient_covariances.get(name)
+        aaT_inv = (
+            cholesky_inverse(
+                cholesky(aaT + self._damping[0] * eye(aaT.shape[0], device=aaT.device))
+            )
+            if aaT is not None
+            else None
```
f-dangel, line 261: Nit: Could invert statement to make it shorter: `None if aaT is None else ...`. Same below.

curvlinops/inverse.py
```diff
@@ -197,3 +204,120 @@ def _matvec(self, x: ndarray) -> ndarray:
...
+                "The input `A` must be an instance of `KFACLinearOperator`."
+            )
+        super().__init__(A.dtype, A.shape)
+        self._A = A
+        self._damping = damping
+        self._cache = cache
+        self._inverse_input_covariances: Dict[str, Tensor] = {}
+        self._inverse_gradient_covariances: Dict[str, Tensor] = {}
+
+    def _compute_or_get_cached_inverse(self, name: str) -> Tuple[Tensor, Tensor]:
```
f-dangel, line 241: Nit: `Tensor` -> `Union[Tensor, None]`.

f-dangel (approved):
Looks good to me, feel free to ignore the nits and merge!

PR #70, opened 2024-02-04
[ADD] Option to specify number of data points

f-dangel:
At the moment, the number of data points is inferred by a pass through the data loader.
This can become a bottleneck on compute clusters when multiple jobs access the same data set in parallel (experience from a recent ICML submission on the Vector cluster).
To resolve this, the PR adds an optional argument `num_data` to all non-experimental linear operators which allows to avoid the data set traversal.

### DIFF
--- a/curvlinops/_base.py
+++ b/curvlinops/_base.py
@@ -84,9 +84,9 @@ class _LinearOperator(LinearOperator):
         self._progressbar = progressbar
 
         self._N_data = (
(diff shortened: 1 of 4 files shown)


runame:
`black` check is currently failing, but besides that, LGTM!

runame (approved):
`black` check is currently failing, but besides that, LGTM!

PR #73, opened 2024-02-05
[REF] Implement `_matmat` instead of `_matvec`

f-dangel:
**Warning:** This is a big refactoring and it will require some iterations to reduce complexity.

Please complain about anything that is incomprehensible, and check for outdated/unclear docstrings.

Currently, all linear operators implement how to multiply with a single vector (`_matvec`).
Multiplying onto more than one vector is simply iterating over the vectors one by one (`_matmat`).
This adds a lot of sweeps through a data set which can be avoided by implementing matrix-matrix,
rather than matrix-vector multiplication.

This PR changes the base class implementation from `_matvec` to `_matmat`.

### DIFF
--- a/curvlinops/_base.py
+++ b/curvlinops/_base.py
@@ -3,21 +3,13 @@
 from typing import Callable, Iterable, List, Optional, Tuple, Union
 from warnings import warn
 
-from backpack.utils.convert_parameters import vector_to_parameter_list
-from numpy import (
-    allclose,
-    argwhere,
-    column_stack,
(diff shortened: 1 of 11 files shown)


f-dangel:
@runame let's first merge #69. I will then refactor `KFACInverseLinearOperator` in this PR, too.

runame (approved):
I might be missing something because I'm not that familiar with the code for the non-KFAC linear operators, but everything looks good to me and the tests are all passing.

PR #76, opened 2024-02-13
[FIX] `Generator` device for sampling

runame:
The device setting during sampling in `MCFisherLinearOperator` and `KFACLinearOperator` is currently broken. This is because `self._generator` will be created with `device="cpu"` to run `_check_deterministic()` and later it is only checked if `self._generator is None`. To fix this, we simply check `if self._generator.device != self._device` as well.

Moreover, the device of `std` was not set explicitly in one place.


f-dangel (approved):
Thanks for catching these!

Issue #77, opened 2024-02-14
[BUG | KFAC] Changing device invalidates parameter mapping


f-dangel:
Whenever `KFAC` is instantiated on GPU, and we call `.to_device(device("cpu))`, this will invalidate the internal mapping between parameter `.data_ptr()`s to module names which is needed to identify a parameter's position in the list format. Currently, this bug can silently pass because we do not check in `matmat` whether all parameter positions are processed.

PR #78, opened 2024-02-14
[FIX] Bug #77: Invalidated parameter mapping after device transfer in KFAC

f-dangel:


runame (approved):
LGTM, thanks for fixing this!

I still wonder if we have to make use of `data_ptr()` or if we can just iterate through the modules and associate module names with parameter names -- they should be unique? In this way there would be no issue when changing the device and the mapping would be more human readable. But there might be an obvious reason why this won't work, haven't thought it through.

f-dangel:
I agree it would be good to refactor this. I will look into using parameter names rather than `.data_ptr()`s as identifiers. If this refactoring is possible, we can remove the checks for outdated `data_ptr`s.

PR #79, opened 2024-02-15
[REF | KFAC] Use module names instead of `data_ptr()`s for mapping

f-dangel:
Manually verified the GPU test works.


runame (approved):
LGTM, I think this is a nice refactor avoiding the previous issue.

PR #80, opened 2024-02-19
[ADD] Squared Frobenius norm estimator

f-dangel:


runame (approved):
LGTM!

Do you think it makes sense add a convenience method that computes an estimate by repeatedly calling `sample()` and averaging, given an argument `num_samples`?

f-dangel:
Yes that might make sense, especially because one could internally compute multiple samples more efficiently via `matmat`'s rather than iterative `matvec`s. 

I'm not sure though if we should return the averaged samples, because we often also want error bars for the stochastic estimates. The individual samples are useful to compute such errorbars.

runame:
Makes sense. We could add a bool argument that makes the method also return a std dev estimate when `True`.

Issue #83, opened 2024-03-07
Improve efficiency of MC Fisher


f-dangel:
Currently, we loop over the data points and MC samples, which is the 'best' way in the sense of FLOPs (`batch_size * mc_samples` backpropagations), but suffers from poor parallelization. We could use `functorch`'s `vmap`, but currently the library relies mostly on `autograd.grad` and I would like to keep it this way as there are certain benefits (e.g. recycling the computation graph when multiplying onto multiple vectors, block-diagonal approximations).

Another way I propose to implement is phrasing multiplication with the MC Fisher as a GGN-vector product where the loss function is such that it's Hessian corresponds to the summed outer product of sampled gradients.
This uses more FLOPs (`batch_size * C` backpropagations), but does not require a loop over `batch_size`. Therefore, this should often yield better run time. One downside of this approach is that it costs as much as multiplying with the exact GGN. This renders the motivation for having an MC-sampled version (less accurate, same cost) weaker.

PR #84, opened 2024-03-07
[REF] Improve efficiency of empirical Fisher

f-dangel:
This PR addresses an efficiency problem in the empirical Fisher, similar to that described in #83 for the MC Fisher.

Remove for-loop over data and use GGN-vector products of a pseudo-loss instead which parallelizes the loop.


runame (approved):
Looks good and is faster, as expected, thanks!

PR #85, opened 2024-03-07
[REF] Remove for-loop over samples in MC Fisher

### LINKED ISSUE #83: Improve efficiency of MC Fisher

f-dangel:
Applies the refactoring described in and resolves #83.

Replaces the expensive `autograd` loops over MC samples and data points via a cheaper internal loop over them (which might be refactored further in the future). Computes a GGNVP with a pseudo-loss whose GGN matches the MC Fisher. This approach costs as much as multiplying with the exact Fisher/GGN, but is more parallel and therefore usually much faster.

@runame please check if this fixes your performance problems before reviewing/merging.

### DIFF
--- a/curvlinops/fisher.py
+++ b/curvlinops/fisher.py
@@ -5,16 +5,17 @@ from __future__ import annotations
 from math import sqrt
 from typing import Callable, Iterable, List, Optional, Tuple, Union
 
+from backpack.hessianfree.ggnvp import ggn_vector_product_from_plist
 from einops import einsum
 from numpy import ndarray
 from torch import (
     Generator,
     Tensor,
     as_tensor,
-    autograd,
     multinomial,
     normal,
     softmax,
+    zeros,
     zeros_like,
 )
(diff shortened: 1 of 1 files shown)


runame (approved):
Thanks!

PR #89, opened 2024-03-12
[REF] Parallelize `sample_grad_output` in MC Fisher

f-dangel:
Parallelizes the loop over data points and MC samples in the method that draws 'would-be' gradients from the model's likelihood. Speeds up `FisherMCLinearOperator`. Please check if the docstrings are up-to-date.


runame (approved):
LGTM!

Issue #71, opened 2024-02-04
Unnecessary device transfer in `_postprocess` method


runame:
In the `_postprocess` method the result of a matrix-vector product will always be [transferred to CPU](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/_base.py#L277). While this is consistent with the scipy interface, in many use cases where we only operate with torch GPU tensors this is not desirable, as it creates unnecessary overhead.

andres-fr:
Chipping in to say that I am almost done writing a library to compute matrix-free SVD/eigh using random matrices, and this was one of my main issues with curvlinops: my library is pytorch "native" and does make use of the Hessian from curvlinops, but not of the scipy routines, so that step was also redundant.

A few thoughts:
1. For most DL scenarios, the overhead is minimal with small matrices. And, in my case, if the matrices are big enough that they cause overhead, they are also big enough not to fit on GPU. So at the end this was a bit of a non-issue for me. This said, there are scenarios where this is definitely an issue.
2. In case it helps someone, I dealt with it by writing a mixing class and inheriting from it. The use case is the following:
  * You have a routine that is pytorch-native
  * You have a curvlinops/scipy linear operator that expects numpy
  * you'd like to make it agnostic to numpy/pytorch input and output
3. At the core of this issue is the fundamental question of, what is curvlinops doing? Originally, it was to enable scipy routines for DL. But as we develop our own routines, this may require reconsidering some fundamental aspects, and potentially redesign.
  * If curvlinops stays true to its origin, the wrapper class I mentioned in point 2 is IMO the way to go. Other 
  * Alternatively, an extra API-layer can be added, where the "routines" are either defined (in pytorch or whatever), or imported (from scipy or whatever). Then there is an API-layer on top, which simply wraps the routines in an agnostic manner, using something like the wrapper I provide here.

Thoughts?

# Snippet

```python
class TorchLinOpWrapper:
    """
    Since this library operates mainly with PyTorch tensors, but some useful
    LinOps interface with NumPy arrays instead, this mixin class acts as a
    wraper on the ``__matmul__`` and ``__rmatmul__`` operators,
    so that the operator expects and returns torch tensors, even when the
    wrapped operator interfaces with NumPy. Usage example::

      # extend NumPy linear operator via multiple inheritance
      class TorchWrappedLinOp(TorchLinOpWrapper, LinOp):
          pass
      lop = TorchWrappedLinOp(...)  # instantiate normally
[...]

PR #91, opened 2024-03-15
Allow for matrix-matrix and matrix-vector products with `KFACLinearOperator` and `KFACInverseLinearOperator` without converting to numpy

runame:
Addresses #71, but only for the `KFACLinearOperator` and `KFACInverseLinearOperator`.

`KFACLinearOperator`/`KFACInverseLinearOperator` is arguably the only linear operator here that is likely to be used for preconditioned-gradient methods for large scale neural networks. Therefore, adding `torch_matmat` and `torch_matvec` methods seem like a simple solution to avoid unnecessary device transfers, which are a bottleneck for this use case. However, this doesn't address the issue in general.

### DIFF
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -355,7 +355,7 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
             parameter, i.e. ``[p1.shape, p2.shape, ...]``. If tensor, has shape ``[D]``.
 
         Raises:
-            ValueError: If the input tensor has the wrong shape.
+            ValueError: If the input tensor has the wrong data type.
         """
         if isinstance(v_torch, list):
             v_torch = [v_torch_i.unsqueeze(0) for v_torch_i in v_torch]
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -1,6 +1,6 @@
 """Implements linear operator inverses."""
 
-from typing import Dict, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Union
 
 from einops import einsum, rearrange
 from numpy import allclose, column_stack, ndarray
@@ -274,19 +274,37 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
             self._inverse_gradient_covariances[name] = ggT_inv
         return aaT_inv, ggT_inv
 
-    def _matmat(self, M: ndarray) -> ndarray:
-        """Multiply a matrix ``M`` x by the inverse of KFAC.
+    def torch_matmat(
+        self, M_torch: Union[Tensor, List[Tensor]], return_tensor: bool = True
+    ) -> Union[Tensor, List[Tensor]]:
+        """Apply the inverse of KFAC to a matrix (multiple vectors) in PyTorch.
+
+        This allows for matrix-matrix products with the inverse KFAC approximation in
+        PyTorch without converting tensors to numpy arrays, which avoids unnecessary
+        device transfers when working with GPUs.
 
         Args:
-             M: Matrix for multiplication.
+            M_torch: Matrix for multiplication. If tensor, has shape ``[D, K]`` with
+                some ``K``.
+            return_tensor: Whether to return the result as a tensor or list of tensors.
 
         Returns:
-             Result of inverse matrix-matrixmultiplication, ``KFAC⁻¹ @ M``.
+            Matrix-multiplication result ``KFAC⁻¹ @ M``. If tensor, has shape ``[D, K]``.
+
+        Raises:
+            ValueError: If the input tensor has the wrong shape.
+            ValueError: If the input tensor's shape is incompatible with the KFAC
+                approximation's shape.
         """
+        if not isinstance(M_torch, list):
+            if M_torch.ndim != 2:
+                raise ValueError(f"expected 2-d tensor, not {M_torch.ndim}-d")
+            if M_torch.shape[0] != self.shape[1]:
+                raise ValueError(f"dimension mismatch: {self.shape}, {M_torch.shape}")
+            M_torch = self._A._torch_preprocess(M_torch)
+
         if not self._A._input_covariances and not self._A._gradient_covariances:
             self._A._compute_kfac()
-
-        M_torch = self._A._preprocess(M)
 
         for mod_name, param_pos in self._A._mapping.items():
             # retrieve the inverses of the Kronecker factors from cache or invert them
@@ -318,4 +336,44 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
                         ggT_inv, M_torch[pos], "i j, m j ... -> m i ..."
                     )
 
+        if return_tensor:
+            M_torch = cat([rearrange(M, "k ... -> (...) k") for M in M_torch], dim=0)
+
+        return M_torch
+
+    def torch_matvec(
+        self, v_torch: Tensor, return_tensor: bool = True
+    ) -> Union[Tensor, List[Tensor]]:
+        """Apply the inverse of KFAC to a vector in PyTorch.
+
+        This allows for matrix-vector products with the inverse KFAC approximation in
+        PyTorch without converting tensors to numpy arrays, which avoids unnecessary
+        device transfers when working with GPUs.
+
+        Args:
+            v_torch: Vector for multiplication. Has shape ``[D]``.
+            return_tensor: Whether to return the result as a tensor or list of tensors.
+
+        Returns:
+            Matrix-multiplication result ``KFAC⁻¹ @ v``. If tensor, has shape ``[D]``.
+
+        Raises:
+            ValueError: If the input tensor has the wrong shape.
+        """
+        M = self.shape[0]
+        if v_torch.shape != (M,) and v_torch.shape != (M, 1):
+            raise ValueError("dimension mismatch")
+        return self.torch_matmat(v_torch.view(-1, 1), return_tensor).squeeze(1)
+
+    def _matmat(self, M: ndarray) -> ndarray:
+        """Apply the inverse of KFAC to a matrix (multiple vectors).
+
+        Args:
+            M: Matrix for multiplication. Has shape ``[D, K]`` with some ``K``.
+
+        Returns:
+            Matrix-multiplication result ``KFAC⁻¹ @ M``. Has shape ``[D, K]``.
+        """
+        M_torch = self._A._preprocess(M)
+        M_torch = self.torch_matmat(M_torch, return_tensor=False)
         return self._A._postprocess(M_torch)
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -275,34 +275,28 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
         return aaT_inv, ggT_inv
 
     def torch_matmat(
-        self, M_torch: Union[Tensor, List[Tensor]], return_tensor: bool = True
+        self, M_torch: Union[Tensor, List[Tensor]]
     ) -> Union[Tensor, List[Tensor]]:
         """Apply the inverse of KFAC to a matrix (multiple vectors) in PyTorch.
 
         This allows for matrix-matrix products with the inverse KFAC approximation in
         PyTorch without converting tensors to numpy arrays, which avoids unnecessary
-        device transfers when working with GPUs.
+        device transfers when working with GPUs and flattening/concatenating.
 
         Args:
-            M_torch: Matrix for multiplication. If tensor, has shape ``[D, K]`` with
-                some ``K``.
-            return_tensor: Whether to return the result as a tensor or list of tensors.
+            M_torch: Matrix for multiplication. If list of tensors, each entry has the
+                same shape as a parameter with an additional leading dimension of size
+                ``K`` for the columns, i.e. ``[(K,) + p1.shape), (K,) + p2.shape, ...]``.
+                If tensor, has shape ``[D, K]`` with some ``K``.
 
         Returns:
-            Matrix-multiplication result ``KFAC⁻¹ @ M``. If tensor, has shape ``[D, K]``.
-
-        Raises:
-            ValueError: If the input tensor has the wrong shape.
-            ValueError: If the input tensor's shape is incompatible with the KFAC
-                approximation's shape.
+            Matrix-multiplication result ``KFAC @ M``. Return type is the same as the
+            type of the input. If list of tensors, each entry has the same shape as a
+            parameter with an additional leading dimension of size ``K`` for the columns,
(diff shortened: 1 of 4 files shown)


curvlinops/kfac.py
```diff
@@ -245,20 +246,67 @@ def to_device(self, device: device):
...
         Args:
-            M: Matrix for multiplication. Has shape ``[D, K]`` with some ``K``.
+            M: Matrix for multiplication. Has shape ``[D, K]`` where ``D`` is the
+                number of parameters, and ``K`` is the number of columns.
 
         Returns:
-            Matrix-multiplication result ``KFAC @ M``. Has shape ``[D, K]``.
+            Matrix in list format. Each entry has the same shape as a parameter with
+            an additional leading dimension of size ``K`` for the columns, i.e.
+            ``[(K,) + p1.shape), (K,) + p2.shape, ...]``.
```
f-dangel, line 259: Is this correct? I could not find a transpose operation in the code that permutes the order from `[D, K]` to `[K, D]`.

curvlinops/kfac.py
```diff
@@ -295,6 +343,46 @@ def _matmat(self, M: ndarray) -> ndarray:
...
 
+        if return_tensor:
+            M_torch = cat([rearrange(M, "k ... -> (...) k") for M in M_torch], dim=0)
+
+        return M_torch
+
+    def torch_matvec(
+        self, v_torch: Tensor, return_tensor: bool = True
+    ) -> Union[Tensor, List[Tensor]]:
+        """Apply KFAC to a vector in PyTorch.
```
f-dangel, line 354: Same comments as for inverse KFAC.

f-dangel (commented):
I made some refactoring proposals.

test/test_kfac.py
```diff
@@ -426,3 +428,64 @@ def test_bug_device_change_invalidates_parameter_mapping():
...
+
+    kfac = KFACLinearOperator(
+        model,
+        loss_func,
+        list(model.parameters()),
+        data,
+        fisher_type="empirical",
+    )
+
+    x = rand(kfac.shape[1], 16, device=dev)
```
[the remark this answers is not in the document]
runame: Just a random number for `num_vectors`, I named this constant to make it more explicit.

curvlinops/inverse.py
```diff
@@ -318,4 +336,44 @@ def _matmat(self, M: ndarray) -> ndarray:
...
+        self, v_torch: Tensor, return_tensor: bool = True
+    ) -> Union[Tensor, List[Tensor]]:
+        """Apply the inverse of KFAC to a vector in PyTorch.
+
+        This allows for matrix-vector products with the inverse KFAC approximation in
+        PyTorch without converting tensors to numpy arrays, which avoids unnecessary
+        device transfers when working with GPUs.
+
+        Args:
+            v_torch: Vector for multiplication. Has shape ``[D]``.
```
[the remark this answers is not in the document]
runame: I mainly introduced this option for `torch_matmat` so that I can use it within `_matmat`. But I think it is actually a cool feature to also support this here, since it will make gradient preconditioning more efficient (could also be done via `torch_matmat`, but I guess it's cleaner to directly support it with `torch_matvec`).

curvlinops/kfac.py
runame: I think the transpose happens [here](https://github.com/f-dangel/curvlinops/blob/kfac-torch-only/curvlinops/kfac.py#L267), or am I missing something?

curvlinops/inverse.py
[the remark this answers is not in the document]
runame: Added support for list inputs for `torch_matvec`.

curvlinops/kfac.py
```diff
@@ -245,20 +245,98 @@ def to_device(self, device: device):
...
+            assert len(M_torch) == len(self._params)
+            for M, p in zip(M_torch, self._params):
+                if len(M) != K:
+                    raise ValueError(
+                        "All input tensors must have the same number of columns."
+                    )
+                if M.shape[1:] != p.shape:
+                    raise ValueError(
+                        "All input tensors must have (K,) + the same shape as the parameters."
+                    )
```
f-dangel, line 304: Refactoring proposal:
```python
column_values = set(M.shape[0] for M in M_torch)
if len(column_values) != 1:
    raise ValueError(f"Number of columns must be equal for all tensors. Got {column_values}")
K = column_values.pop()
for M, p in zip(M_torch, self._params):
    if M.shape != (K,) + p.shape:
        raise ValueError("All input tensors must have (K, ) + p.shape. Got {M.shape} versus {(K,) + p.shape}")
```

curvlinops/kfac.py
```diff
@@ -295,6 +373,54 @@ def _matmat(self, M: ndarray) -> ndarray:
...
+            ValueError: If the input tensor has the wrong shape.
+        """
+        if isinstance(v_torch, list):
+            v_torch = [v_torch_i.unsqueeze(0) for v_torch_i in v_torch]
+            result = self.torch_matmat(v_torch)
+            return [res.squeeze(0) for res in result]
+        else:
+            M = self.shape[0]
+            if v_torch.shape != (M,):
+                raise ValueError("The input vector has the wrong shape.")
```
f-dangel, line 410: This should be caught inside `torch_matmat` already and could be removed.

curvlinops/kfac.py
```diff
@@ -295,6 +373,54 @@ def _matmat(self, M: ndarray) -> ndarray:
...
+            return self.torch_matmat(v_torch.unsqueeze(1)).squeeze(1)
```
f-dangel, line 411: If we remove the exception above, we should change the `1`s into `-1`s.

curvlinops/inverse.py
```diff
@@ -318,4 +330,52 @@ def _matmat(self, M: ndarray) -> ndarray:
...
+        """
+        if isinstance(v_torch, list):
+            v_torch = [v_torch_i.unsqueeze(0) for v_torch_i in v_torch]
+            result = self.torch_matmat(v_torch)
+            return [res.squeeze(0) for res in result]
+        else:
+            M = self.shape[0]
+            if v_torch.shape != (M,):
+                raise ValueError("The input vector has the wrong shape.")
+            return self.torch_matmat(v_torch.unsqueeze(1)).squeeze(1)
```
f-dangel, line 368: I actually think we can remove the exception because it should be caught in `torch_matmat`. Then we would have to replace the `1`s by `-1`s.

curvlinops/kfac.py
f-dangel, line 410: maybe a meaningful check would be here to check that `v_torch` is indeed a `Tensor`, and not something like a `Tuple[Tensor]`.

test/test_inverse.py
```diff
@@ -185,3 +185,85 @@ def test_KFAC_inverse_damped_matmat(
...
+        data,
+        loss_average=loss_average,
+    )
+    inv_KFAC = KFACInverseLinearOperator(KFAC, damping=(delta, delta))
+
+    num_vectors = 2
+    X = torch.rand(KFAC.shape[1], num_vectors)
+    inv_KFAC_X = inv_KFAC.torch_matmat(X)
+    assert inv_KFAC_X.dtype == X.dtype
+    assert inv_KFAC_X.device == X.device
```
f-dangel, line 208: This should fail on GPU atm because `torch_matmat` does not move `X` to another device and `X` is always created on CPU.

test/test_inverse.py
```diff
@@ -185,3 +185,85 @@ def test_KFAC_inverse_damped_matmat(
...
+        params,
+        data,
+        loss_average=loss_average,
+    )
+    inv_KFAC = KFACInverseLinearOperator(KFAC, damping=(delta, delta))
+
+    x = torch.rand(KFAC.shape[1])
+    inv_KFAC_x = inv_KFAC.torch_matvec(x)
+    assert inv_KFAC_x.dtype == x.dtype
+    assert inv_KFAC_x.device == x.device
```
f-dangel, line 247: Same as above.

test/test_inverse.py
```diff
@@ -185,3 +185,85 @@ def test_KFAC_inverse_damped_matmat(
...
+    # Test list input format
+    # split parameter blocks
+    dims = [p.numel() for p in KFAC._params]
+    split_x = x.split(dims)
+    # unflatten parameter dimension
+    assert len(split_x) == len(KFAC._params)
+    x_list = [res.reshape(p.shape) for res, p in zip(split_x, KFAC._params)]
+    inv_kfac_x_list = inv_KFAC.torch_matvec(x_list)
+    inv_kfac_x_list = torch.cat(
+        [rearrange(M, "... -> (...)") for M in inv_kfac_x_list], dim=0
```
f-dangel, line 259: Nit: Could remove `dim=0` as it is the default.

test/test_inverse.py
```diff
@@ -185,3 +185,85 @@ def test_KFAC_inverse_damped_matmat(
...
+    assert inv_KFAC_X.dtype == X.dtype
+    assert inv_KFAC_X.device == X.device
+    assert inv_KFAC_X.shape == (KFAC.shape[0], num_vectors)
+    inv_KFAC_X = inv_KFAC_X.cpu().numpy()
+
+    # Test list input format
+    x_list = KFAC._torch_preprocess(X)
+    inv_KFAC_x_list = inv_KFAC.torch_matmat(x_list)
+    inv_KFAC_x_list = torch.cat(
+        [rearrange(M, "k ... -> (...) k") for M in inv_KFAC_x_list], dim=0
```
f-dangel, line 216: Nit: Could remove `dim=0` as it is the default.

test/test_kfac.py
```diff
@@ -426,3 +429,88 @@ def test_bug_device_change_invalidates_parameter_mapping():
     kfac_x_cpu = kfac @ x
 
     report_nonclose(kfac_x_gpu, kfac_x_cpu)
+
+
+@mark.parametrize("dev", DEVICES, ids=DEVICES_IDS)
+def test_torch_matmat(dev: device):
```
f-dangel, line 435: Can we use `cases` just like for the inverse KFAC tests?

test/test_kfac.py
```diff
@@ -426,3 +429,88 @@ def test_bug_device_change_invalidates_parameter_mapping():
...
+        loss_func,
+        list(model.parameters()),
+        data,
+        fisher_type="empirical",
+    )
+
+    num_vectors = 16
+    x = rand(kfac.shape[1], num_vectors, device=dev)
+    kfac_x = kfac.torch_matmat(x)
+    assert x.device == kfac_x.device
```
f-dangel, line 454: This should error on GPU (same as above).

test/test_kfac.py
```diff
@@ -426,3 +429,88 @@ def test_bug_device_change_invalidates_parameter_mapping():
...
+    assert x.dtype == kfac_x.dtype
+    assert kfac_x.shape == (kfac.shape[0], x.shape[1])
+    kfac_x = kfac_x.cpu().numpy()
+
+    # Test list input format
+    x_list = kfac._torch_preprocess(x)
+    kfac_x_list = kfac.torch_matmat(x_list)
+    kfac_x_list = cat([rearrange(M, "k ... -> (...) k") for M in kfac_x_list], dim=0)
```
f-dangel, line 462: Nit: Could remove `dim=0` as it is the default.

test/test_kfac.py
```diff
@@ -426,3 +429,88 @@ def test_bug_device_change_invalidates_parameter_mapping():
...
+    report_nonclose(kfac_x, kfac_x_list.cpu().numpy())
+
+    # Test against multiplication with dense matrix
+    kfac_mat = kfac.torch_matmat(torch_eye(kfac.shape[1], device=dev))
+    kfac_mat_x = kfac_mat @ x
+    report_nonclose(kfac_x, kfac_mat_x.cpu().numpy())
+
+    # Test against _matmat
+    kfac_x_numpy = kfac @ x.cpu().numpy()
+    report_nonclose(kfac_x, kfac_x_numpy)
```
f-dangel, line 516: Same comments as before.

f-dangel (commented):
Minor things. After this round I believe we are ready to merge, maybe up to some more minor nits.

test/test_inverse.py
runame: True. Should `case` not also return the device, so that we can ensure that all operations are executed on the desired device?

test/test_kfac.py
runame: This doesn't error here because I don't use `cases`.

test/test_inverse.py
```diff
@@ -185,3 +185,83 @@ def test_KFAC_inverse_damped_matmat(
...
+    # Test list input format
+    x_list = KFAC._torch_preprocess(X)
+    inv_KFAC_x_list = inv_KFAC.torch_matmat(x_list)
+    inv_KFAC_x_list = torch.cat(
+        [rearrange(M, "k ... -> (...) k") for M in inv_KFAC_x_list]
+    )
+    report_nonclose(inv_KFAC_X, inv_KFAC_x_list.cpu().numpy())
+
+    # Test against multiplication with dense matrix
+    inv_KFAC_mat = inv_KFAC.torch_matmat(torch.eye(inv_KFAC.shape[1]))
```
f-dangel, line 221: Should specify device and dtype here.

test/test_inverse.py
```diff
@@ -185,3 +185,83 @@ def test_KFAC_inverse_damped_matmat(
...
+    split_x = x.split(dims)
+    # unflatten parameter dimension
+    assert len(split_x) == len(KFAC._params)
+    x_list = [res.reshape(p.shape) for res, p in zip(split_x, KFAC._params)]
+    inv_kfac_x_list = inv_KFAC.torch_matvec(x_list)
+    inv_kfac_x_list = torch.cat([rearrange(M, "... -> (...)") for M in inv_kfac_x_list])
+    report_nonclose(inv_KFAC_x, inv_kfac_x_list.cpu().numpy())
+
+    # Test against multiplication with dense matrix
+    inv_KFAC_mat = inv_KFAC.torch_matmat(torch.eye(inv_KFAC.shape[1]))
```
f-dangel, line 262: Should specify device and dtype here.

test/test_kfac.py
f-dangel: Leaving this open for discussion. I believe it would make the code shorter and also allow testing different `dtypes` in the long run (which atm we are not).

f-dangel (commented):
Found some more minor things, but looks great! One more thing to discuss is whether we want to switch to using `cases` in the tests of KFAC.

test/test_inverse.py
runame: Yea I think we can do this instead of explicitly returning `device`. We can't use `inv_KFAC.dtype` though because I think this will be a numpy dtype.

f-dangel (approved):
Looks ready to me!

PR #93, opened 2024-03-16
Add heuristic and exact damping to `KFACInverseLinearOperator`

### LINKED ISSUE #92: Add option for heuristic and exact damping to `KFACInverseLinearOperator`
There are two different ways to set the damping for the `KFACInverseLinearOperator` that we should implement:

1. 'Heuristic' damping, introduced in section 6.3 in the [original K-FAC paper](https://arxiv.org/abs/1503.05671).
2. 'Exact' damping, which can be efficiently implemented for Kronecker factored matrices, e.g. see equation (21) in [Grosse et al., 2023](https://arxiv.org/abs/2308.03296).

There is also an 'adaptive' damping scheme, as described in the [original K-FAC paper](https://arxiv.org/abs/1503.05671), but I do not plan to implement this for now.

runame:
Resolves #92.

~Please review and merge #91, #95, and #98 first.~

### DIFF
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -221,18 +221,30 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
         use_exact_damping: bool = False,
         cache: bool = True,
     ):
-        """Store the linear operator whose inverse should be represented.
+        r"""Store the linear operator whose inverse should be represented.
 
         Args:
             A: ``KFACLinearOperator`` whose inverse is formed.
-            damping: Damping value(s) for all input and gradient covariances.
-                Default: ``0.``.
-            use_heuristic_damping: Whether to use a heuristic damping strategy.
+            damping: Damping value(s) for all input and gradient covariances. If tuple,
+                the first value is used for the input covariances and the second value
+                for the gradient covariances. Note that if heuristic or exact damping is
+                used the damping cannot be a tuple. Default: ``0.``.
+            use_heuristic_damping: Whether to use a heuristic damping strategy by
+                `Martens and Grosse, 2015 <https://arxiv.org/abs/1503.05671>`_
+                (Section 6.3). For input covariances
+                :math:`A \in \mathbb{R}^{n \times n}` and gradient covariances
+                :math:`B \in \mathbb{R}^{m \times m}`, we define
+                :math:`\pi := \sqrt{m \frac{\text{tr}(A)}{n \text{tr}(B)}}` and set the
+                damping for the input covariances :math:`A` to
+                :math:`\max(\sqrt{\text{damping}} \pi, \text{min_damping})` and for the
+                gradient covariances :math:`B` to
+                :math:`\max(\sqrt{\text{damping}} / \pi, \text{min_damping})`.
                 Default: ``False``.
-            min_damping: Minimum damping value. Default: ``1e-8``. Only used if
-                ``use_heuristic_damping`` is ``True``.
-            use_exact_damping: Whether to use exact damping.
-                Default: ``False``.
+            min_damping: Minimum damping value. Only used if ``use_heuristic_damping``
+                is ``True``. Default: ``1e-8``.
+            use_exact_damping: Whether to use exact damping, i.e. to invert
+                :math:`(A \otimes B) + \text{damping} \mathbf{I}`. This is implemented
+                via eigendecompositions of the Kronecker factors. Default: ``False``.
             cache: Whether to cache the inverses of the Kronecker factors.
                 Default: ``True``.
 
@@ -276,6 +288,7 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
             Damping values for the input and gradient covariances.
         """
         if self._use_heuristic_damping and aaT is not None and ggT is not None:
+            # Martens and Grosse, 2015 (https://arxiv.org/abs/1503.05671) (Section 6.3)
             aaT_eig_mean = aaT.trace() / len(aaT)
             ggT_eig_mean = ggT.trace() / len(ggT)
             if aaT_eig_mean >= 0.0 and ggT_eig_mean > 0.0:
@@ -301,7 +314,8 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
 
         Returns:
             Tuple of inverses (or eigendecompositions) of the input and gradient
-            covariance Kronecker factors.
+            covariance Kronecker factors. Can be ``None`` if the input or gradient
+            covariance is ``None`` (e.g. for a bias or if weights are excluded).
         """
         if self._use_exact_damping:
             # Compute eigendecomposition to perform damped preconditioning in
@@ -376,10 +390,13 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
             # https://arxiv.org/abs/2308.03296.
             aaT_eigvecs, aaT_eigvals = aaT_inv
             ggT_eigvecs, ggT_eigvals = ggT_inv
+            # Transform in eigenbasis.
             M_joint = einsum(
                 ggT_eigvecs, M_joint, aaT_eigvecs, "i j, m i k, k l -> m j l"
             )
+            # Divide by damped eigenvalues to perform the inversion.
             M_joint.div_(outer(ggT_eigvals, aaT_eigvals).add_(self._damping))
+            # Transform back to standard basis.
             M_joint = einsum(
                 ggT_eigvecs, M_joint, aaT_eigvecs, "i j, m j k, l k -> m i l"
             )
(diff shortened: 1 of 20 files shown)


test/test_inverse.py
```diff
@@ -177,10 +181,170 @@ def test_KFAC_inverse_damped_matmat(
...
+        params,
+        data,
+        loss_average=loss_average,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+
+    # manual exactly damped inverse
+    inv_KFAC_naive = torch.inverse(
+        KFAC.torch_matmat(torch.eye(KFAC.shape[0])) + delta * torch.eye(KFAC.shape[0])
+    )
```
f-dangel, line 332: Is there a reason to use `torch_matmat` over the public `matmat` with a `numpy` array?

f-dangel: I am confused because in L346 `X` is a `numpy` array, but `inv_KFAC_naive` seems to be a `torch.Tensor` that is multiplied onto it. I expect this should crash if we execute the tests on a machine with CUDA device.

f-dangel (commented):
My comments are mostly about adding documentation in the right places.
I noticed the code is becoming increasingly complex. If we plan to include further strategies, we might have to think about refactorings.

test/test_inverse.py
```diff
@@ -177,10 +181,170 @@ def test_KFAC_inverse_damped_matmat(
...
+    inv_KFAC = KFACInverseLinearOperator(KFAC, damping=delta, cache=cache)
+    # and as a tuple
+    inv_KFAC_tuple = KFACInverseLinearOperator(
+        KFAC, damping=(delta, delta), cache=cache
+    )
+
+    num_vectors = 2
+    X = random.rand(KFAC.shape[1], num_vectors)
+    report_nonclose(inv_KFAC @ X, inv_KFAC_naive @ X, rtol=5e-2)
+    report_nonclose(inv_KFAC_tuple @ X, inv_KFAC @ X, rtol=5e-2)
```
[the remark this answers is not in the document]
runame: I was also confused by this, it might be the numerical instability of the inversion? I think I might have converted everything to `float64` to verify that it requires less `rtol`, but not 100% sure anymore. I think most of the tests comparing for exact equivalence would benefit from being performed in `float64`.

curvlinops/inverse.py
```diff
@@ -212,68 +215,234 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
...
+        Returns:
+            Tuple of inverses (or eigendecompositions) of the input and gradient
+            covariance Kronecker factors.
+        """
+        if self._use_exact_damping:
+            # Compute eigendecomposition to perform damped preconditioning in
+            # Kronecker-factored eigenbasis (KFE).
+            aaT_eigvals, aaT_eigvecs = (None, None) if aaT is None else eigh(aaT)
+            ggT_eigvals, ggT_eigvecs = (None, None) if ggT is None else eigh(ggT)
+            return (aaT_eigvecs, aaT_eigvals), (ggT_eigvecs, ggT_eigvals)
```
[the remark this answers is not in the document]
runame: It's not, see new comments in `torch_matmat`.

curvlinops/inverse.py
```diff
@@ -212,68 +215,234 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
...
+            aaT: Input covariance matrix.
+            ggT: Gradient covariance matrix.
+
+        Returns:
+            Damping values for the input and gradient covariances.
+        """
+        if self._use_heuristic_damping and aaT is not None and ggT is not None:
+            aaT_eig_mean = aaT.trace() / len(aaT)
+            ggT_eig_mean = ggT.trace() / len(ggT)
+            if aaT_eig_mean >= 0.0 and ggT_eig_mean > 0.0:
```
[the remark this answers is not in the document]
runame: If either `aaT` or `ggT` are `None` we can't compute the damping heuristic and fall back to just adding the `damping` directly (i.e. `\pi=1`).

test/test_inverse.py
```diff
@@ -177,10 +181,170 @@ def test_KFAC_inverse_damped_matmat(
...
+
+    # use exact damping with KFACInverseLinearOperator
+    inv_KFAC = KFACInverseLinearOperator(
+        KFAC, damping=delta, cache=cache, use_exact_damping=True
+    )
 
     num_vectors = 2
     X = random.rand(KFAC.shape[1], num_vectors)
+    # test for equivalence
     report_nonclose(inv_KFAC @ X, inv_KFAC_naive @ X, rtol=5e-2)
```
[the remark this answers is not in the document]
runame: I assume not, I usually start from not specifying `rtol` at all.

test/test_inverse.py
runame: True, I added `.cpu().numpy()` in some places. And there is no reason for using `torch_matmat` besides that I computed the inverse in `PyTorch`.

curvlinops/inverse.py
```diff
@@ -212,68 +215,255 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
...
+                damping for the input covariances :math:`A` to
+                :math:`\max(\sqrt{\text{damping}} \pi, \text{min_damping})` and for the
+                gradient covariances :math:`B` to
+                :math:`\max(\sqrt{\text{damping}} / \pi, \text{min_damping})`.
+                Default: ``False``.
+            min_damping: Minimum damping value. Only used if ``use_heuristic_damping``
+                is ``True``. Default: ``1e-8``.
+            use_exact_damping: Whether to use exact damping, i.e. to invert
+                :math:`(A \otimes B) + \text{damping} \mathbf{I}`. This is implemented
+                via eigendecompositions of the Kronecker factors. Default: ``False``.
```
f-dangel, line 247: Maybe add reference to influence function paper here.

f-dangel: Also, can we add `KFACInverseLinearOperator` to the docs? Maybe under **Inverses**?

curvlinops/inverse.py
```diff
@@ -212,68 +215,255 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
...
+        self._inverse_gradient_covariances: Dict[str, KFAC_INV_TYPE] = {}
+
+    def _compute_damping(
+        self, aaT: Optional[Tensor], ggT: Optional[Tensor]
+    ) -> Tuple[float, float]:
+        """Compute the damping values for the input and gradient covariances.
+
+        Args:
+            aaT: Input covariance matrix.
+            ggT: Gradient covariance matrix.
```
f-dangel, line 285: Can we add what it means for an argument to be `None`?

curvlinops/inverse.py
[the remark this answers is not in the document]
f-dangel: I see, closing this comment but I've added a new comment where to document what it means for `aaT` or `ggT` to be `None`.

test/test_inverse.py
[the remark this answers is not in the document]
f-dangel: Then, let's run this test in `float64` and hope we can get an `atol` which is a factor of 10 smaller.

test/test_inverse.py
```diff
@@ -170,17 +174,187 @@ def test_KFAC_inverse_damped_matmat(
...
+@mark.parametrize(
+    "exclude", [None, "weight", "bias"], ids=["all", "no_weights", "no_biases"]
+)
+@mark.parametrize(
+    "separate_weight_and_bias", [True, False], ids=["separate_bias", "joint_bias"]
+)
+def test_KFAC_inverse_heuristically_damped_matmat(  # noqa: C901
+    case, cache: bool, exclude: str, separate_weight_and_bias: bool, delta: float = 1e-2
+):
+    """Test matrix-matrix multiplication by an inverse (heuristically) damped KFAC
```
f-dangel, line 220: I would also vote to run this test in `float64` to be able to reduce `rtol`.

test/test_inverse.py
```diff
@@ -170,17 +174,187 @@ def test_KFAC_inverse_damped_matmat(
...
+    "exclude", [None, "weight", "bias"], ids=["all", "no_weights", "no_biases"]
+)
+@mark.parametrize(
+    "separate_weight_and_bias", [True, False], ids=["separate_bias", "joint_bias"]
+)
+def test_KFAC_inverse_exactly_damped_matmat(
+    case, cache: bool, exclude: str, separate_weight_and_bias: bool, delta: float = 1e-2
+):
+    """Test matrix-matrix multiplication by an inverse (exactly) damped KFAC approximation."""
+    model_func, loss_func, params, data = case
```
f-dangel, line 318: Also vote for casting to `float64` and lower `rtol`

f-dangel (approved):
Feel free to address and merge! I would vote though to switch the damped KFAC inverse tests to using `float64` so we can lower the tolerance (hopefully by 2, but at least by one order of magnitude).

PR #95, opened 2024-03-20
Add `trace`, `det`, `logdet`, and `frobenius_norm` property to `KFACLinearOperator`

### LINKED ISSUE #96: Add `trace`, `det`, `logdet`, and `frobenius_norm` properties to `KFACLinearOperator`
Since the `KFACLinearOperator` stores its state explicitly and the Kronecker product has many convenient properties, we can efficiently compute the trace, determinant, log determinant, and Frobenius norm. This is much more efficient than computing it naively by building the dense matrix or using a stochastic trace estimator.

runame:
Resolves #96.

Please merge #91 before reviewing this PR.

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -379,7 +379,7 @@ class KFACLinearOperator(_LinearOperator):
                     )
 
         if return_tensor:
-            M_torch = cat([rearrange(M, "k ... -> (...) k") for M in M_torch], dim=0)
+            M_torch = cat([rearrange(M, "k ... -> (...) k") for M in M_torch])
 
         return M_torch
 
@@ -403,7 +403,7 @@ class KFACLinearOperator(_LinearOperator):
             parameter, i.e. ``[p1.shape, p2.shape, ...]``. If tensor, has shape ``[D]``.
 
         Raises:
-            ValueError: If the input tensor has the wrong shape.
+            ValueError: If the input tensor has the wrong data type.
         """
         if isinstance(v_torch, list):
             v_torch = [v_torch_i.unsqueeze(0) for v_torch_i in v_torch]
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -886,7 +886,7 @@ class KFACLinearOperator(_LinearOperator):
             else:
                 for p_name in param_pos.keys():
                     n = (
-                        self._input_covariances[mod_name][0]
+                        self._input_covariances[mod_name].shape[0]
                         if p_name == "weight"
                         else 1
                     )
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -255,7 +255,7 @@ class KFACLinearOperator(_LinearOperator):
         Returns:
             Matrix in list format. Each entry has the same shape as a parameter with
             an additional leading dimension of size ``K`` for the columns, i.e.
-            ``[(K,) + p1.shape), (K,) + p2.shape, ...]``.
+            ``[(K,) + p1.shape, (K,) + p2.shape, ...]``.
         """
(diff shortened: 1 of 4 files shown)


f-dangel (commented):
One question upfront: Is there any specific benefit of implementing these properties through `@property` other than them being cached? Why not implement them through getters? A benefit would be that we don't have to reset any attributes. A downside would be that there is no caching mechanism. Am I missing something else?

runame:
No, the main reason for using properties here is the caching and that it fits semantically (they are attributes/properties of the matrix). I think the resetting is a small price to pay for being able to cache the result, especially since we just have to add one new line in the resetting method for every additional property we implement, so seems easy to extend.

curvlinops/kfac.py
```diff
@@ -757,3 +769,161 @@ def compute_parameter_mapping(
...
+                ):
+                    squared_frob_aaT = self._input_covariances[mod_name].square().sum()
+                    self._frobenius_norm += squared_frob_aaT * squared_frob_ggT
+                else:
+                    for p_name in param_pos.keys():
+                        squared_frob_mod = squared_frob_ggT.clone()
+                        if p_name == "weight":
+                            squared_frob_mod *= (
+                                self._input_covariances[mod_name].square().sum()
+                            )
```
f-dangel, line 925: Refactoring proposal:
```python
squared_frob = squared_frob_ggT * (self._input_covariances[mod_name].square().sum() if p_name == "weight" else 1)
self._frobenius_norm += squared_frob_mod
```

test/test_kfac.py
```diff
@@ -519,3 +520,200 @@ def test_torch_matvec(case):
...
+            torch_eye(ggT.shape[0], dtype=ggT.dtype, device=ggT.device), alpha=delta
+        )
+
+    # Check for equivalence of the logdet property and naive log determinant computation
+    log_det = kfac.logdet
+    # verify that the log determinant is finite and not nan
+    assert not isinf(log_det) and not isnan(log_det)
+    sign, logabsdet = slogdet(kfac @ eye(kfac.shape[1]))
+    det_through_logdet_naive = sign * exp(logabsdet)
+    report_nonclose(exp(log_det.cpu().numpy()), det_through_logdet_naive)
```
f-dangel, line 714: Q: Why not compare `log_det` with `sign * logabsdet`?

f-dangel (approved):
Had some refactoring suggestions, but looks good already!

test/test_kfac.py
runame: Was numerically less similar, but I just increased the damping and now compare the log dets directly.

curvlinops/kfac.py
```diff
@@ -757,3 +769,161 @@ def compute_parameter_mapping(
...
+            self._logdet = 0.0
+            for mod_name, param_pos in self._mapping.items():
+                m = len(self._gradient_covariances[mod_name])
+                logdet_ggT = self._gradient_covariances[mod_name].logdet()
+                if (
+                    not self._separate_weight_and_bias
+                    and "weight" in param_pos.keys()
+                    and "bias" in param_pos.keys()
+                ):
+                    n = len(self._input_covariances[mod_name])
```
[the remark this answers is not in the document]
runame: I recently decided to use `len` instead of `shape[0]` because it is *slightly* [faster for numpy](https://stackoverflow.com/questions/51332056/numpy-performance-gap-between-lenarr-and-arr-shape0) and since everything is typed anyway, it is usually clear what types objects have. But I just checked and for torch `shape[0]` seems to be slightly faster, so I switched to this -- and I agree that it is making it even more explicit that it is a tensor object.
Not that the performance of this matters at all, but had to somehow choose what to use haha

PR #99, opened 2024-03-27
[ADD] Support `BCEWithLogitsLoss` in Fisher-MC and KFAC

### LINKED ISSUE #90: Support `BCEWithLogitsLoss`
Requested by @wiseodd for `laplace-torch`.

This consists of three parts:

- [x] Support `sqrt_hessian` for `KFAC` (type 2)
- [x] Support `sample_grad_output` for `FisherMC`
- [x] Support `draw_label` for `KFAC`  (type 1)

f-dangel:
This makes `curvlinops` support binary classification. Resolves #90

### DIFF
--- a/curvlinops/fisher.py
+++ b/curvlinops/fisher.py
@@ -8,17 +8,8 @@ from typing import Callable, Iterable, List, Optional, Tuple, Union
 from backpack.hessianfree.ggnvp import ggn_vector_product_from_plist
 from einops import einsum, rearrange
 from numpy import ndarray
-from torch import (
-    Generator,
-    Tensor,
-    as_tensor,
-    multinomial,
-    normal,
-    softmax,
-    zeros,
-    zeros_like,
-)
-from torch.nn import CrossEntropyLoss, MSELoss, Parameter
+from torch import Generator, Tensor, as_tensor, normal, softmax, zeros, zeros_like
+from torch.nn import BCEWithLogitsLoss, CrossEntropyLoss, MSELoss, Parameter
 from torch.nn.functional import one_hot
 
 from curvlinops._base import _LinearOperator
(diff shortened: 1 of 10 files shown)


curvlinops/fisher.py
```diff
@@ -253,6 +244,7 @@ def sample_grad_output(self, output: Tensor, num_samples: int) -> Tensor:
             output: model prediction ``f`` for multiple data with batch axis as
                 0th dimension.
             num_samples: Number of samples to draw.
+            y: Ground truth.
```
runame, line 247: I think I would prefer another term here, something like "empirical label" or "label from data". "Ground truth" implies to me that this is what we are trying to approximate or what would be the ideal object to use.

curvlinops/kfac.py
```diff
@@ -555,6 +563,18 @@ def _compute_loss_and_backward(self, output: Tensor, y: Tensor):
...
+
+                if (
+                    isinstance(self._loss_func, (BCEWithLogitsLoss, MSELoss))
+                    and self._loss_func.reduction == "mean"
+                ):
+                    # ``BCEWithLogitsLoss`` and ``MSELoss`` also average over non-batch
+                    # dimensions. We have to scale the loss to incorporate this scaling
+                    # as we cannot generally achieve it by incorporating it into the
+                    # drawn sample.
+                    C = output.shape[1:].numel()
```
runame, line 575: Should `output.shape[1]` be sufficient here since we always reshape the outputs to be 2d?

runame (approved):
Looks good to me!

curvlinops/fisher.py
f-dangel: Agree.

curvlinops/kfac.py
f-dangel: Made more explicit.

PR #102, opened 2024-04-17
[ADD] Option to retry failed Cholesky decompositions in double precision in `KFACInverseLinearOperator`

### LINKED ISSUE #101: Numerical stability of `cholesky` in `KFACInverseLinearOperator`
The Cholesky decomposition used for inversion in `KFACInverseLinearOperator` is sensitive to numerical stability. Therefore, it might make sense to add an option to retry failed decompositions in double precision, similar to [here](https://github.com/facebookresearch/optimizers/blob/main/matrix_functions.py#L212-L223).

runame:
Resolves #101.

### DIFF
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -324,7 +324,7 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
             Tuple of inverses (or eigendecompositions) of the input and gradient
             covariance Kronecker factors. Can be ``None`` if the input or gradient
             covariance is ``None`` (e.g. the input covariances for biases).
-        
+
         Raises:
             Exception: If a Cholesky decomposition (and optionally the retry in double
                 precision) fails.
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -2,11 +2,12 @@
 
 from math import sqrt
 from typing import Dict, List, Optional, Tuple, Union
+from warnings import warn
 
 from einops import einsum, rearrange
 from numpy import allclose, column_stack, ndarray
 from scipy.sparse.linalg import LinearOperator, cg
-from torch import Tensor, cat, cholesky_inverse, eye, outer
+from torch import Tensor, cat, cholesky_inverse, eye, float64, outer
 from torch.linalg import cholesky, eigh
 
 from curvlinops.kfac import KFACLinearOperator
@@ -220,6 +221,7 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
         min_damping: float = 1e-8,
         use_exact_damping: bool = False,
         cache: bool = True,
+        retry_double_precision: bool = True,
     ):
         r"""Store the linear operator whose inverse should be represented.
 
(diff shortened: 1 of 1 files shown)


curvlinops/inverse.py
```diff
@@ -327,30 +337,69 @@ def _compute_inverse_factors(
             return (aaT_eigvecs, aaT_eigvals), (ggT_eigvecs, ggT_eigvals)
         else:
             damping_aaT, damping_ggT = self._compute_damping(aaT, ggT)
-            aaT_inv = (
-                None
-                if aaT is None
-                else cholesky_inverse(
-                    cholesky(
+
+            # Compute inverse of aaT via Cholesky decomposition
```
f-dangel, line 341: Could you extract
```
aaT_damped = aaT.add(
                            eye(aaT.shape[0], dtype=aaT.dtype, device=aaT.device),
                            alpha=damping_aaT,
                        )
```
outside the try/except block to reduce duplication?

curvlinops/inverse.py
```diff
@@ -327,30 +337,69 @@ def _compute_inverse_factors(
...
                         )
                     )
                 )
-            )
-            ggT_inv = (
-                None
-                if ggT is None
-                else cholesky_inverse(
-                    cholesky(
+            except Exception as exception:
```
f-dangel, line 353: Should we specify the exception type further to make sure the re-try happens only if cholesky crashes due non-convergence?

curvlinops/inverse.py
```diff
@@ -327,30 +337,69 @@ def _compute_inverse_factors(
...
+                    aaT = aaT.to(float64)
+                    aaT_chol = cholesky(
+                        aaT.add(
+                            eye(aaT.shape[0], dtype=aaT.dtype, device=aaT.device),
+                            alpha=damping_aaT,
+                        )
+                    )
+                else:
+                    raise exception
+            aaT_inv = None if aaT_chol is None else cholesky_inverse(aaT_chol)
```
f-dangel, line 370: Should convert back `aaT_inv` or `aaT_chol` to the original dtype if the re-try was triggered.

curvlinops/inverse.py
```diff
@@ -327,30 +337,69 @@ def _compute_inverse_factors(
...
+                    ggT = ggT.to(float64)
+                    ggT_chol = cholesky(
+                        ggT.add(
+                            eye(ggT.shape[0], dtype=ggT.dtype, device=ggT.device),
+                            alpha=damping_ggT,
+                        )
+                    )
+                else:
+                    raise exception
+            ggT_inv = None if ggT_chol is None else cholesky_inverse(ggT_chol)
```
f-dangel, line 401: See comments on the first inversion.

f-dangel (approved):
Some smaller things that need to be addressed. Feel free to fix and merge.

curvlinops/inverse.py
runame: Let's go for `RuntimeError`

PR #104, opened 2024-04-19
[REF] Make KFAC computation side effect free

f-dangel:
Before this PR, we call `.backward()` on a tensor to trigger the backward hooks that compute the grad-output-based Kronecker factors of KFAC. 
As a side effect, this will compute `.grad` of all leafs of the tensor we are differentiating.
This is problematic for use cases in optimization that use the `.grad` attribute, which is modified though by creating a `KFACLinearOperator`. 
This PR changes the `.backward()` calls into `autograd.grad(...)` which does not modify any `.grad` attributes and therefore makes the computation of KFAC side effect free.

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -513,7 +513,6 @@ class KFACLinearOperator(_LinearOperator):
             self._compute_loss_and_backward(output, y)
 
         # clean up
-        self._model_func.zero_grad()
         for handle in hook_handles:
             handle.remove()
 
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -537,10 +537,6 @@ class KFACLinearOperator(_LinearOperator):
             output = rearrange(output, "batch ... c -> (batch ...) c")
             y = rearrange(y, "batch ... c -> (batch ...) c")
 
-        # we will differentiate w.r.t. the model parameters to trigger the backward
-        # hooks that compute the grad-output based Kronecker factors
-        params = [p for p in self._model_func.parameters() if p.requires_grad]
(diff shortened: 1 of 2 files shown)


curvlinops/kfac.py
```diff
@@ -536,6 +537,10 @@ def _compute_loss_and_backward(self, output: Tensor, y: Tensor):
             output = rearrange(output, "batch ... c -> (batch ...) c")
             y = rearrange(y, "batch ... c -> (batch ...) c")
 
+        # we will differentiate w.r.t. the model parameters to trigger the backward
+        # hooks that compute the grad-output based Kronecker factors
+        params = [p for p in self._model_func.parameters() if p.requires_grad]
```
runame, line 542: What is the reason not to use `self._params` here?

runame (approved):
Looks good! Maybe you could add a unit test to avoid regressing to the same behaviour, i.e. simply check that the KFAC computation does not modify the `.grad` attribute of the model parameters.

curvlinops/kfac.py
f-dangel: Yap, that's much better!

f-dangel:
> Looks good! Maybe you could add a unit test to avoid regressing to the same behaviour, i.e. simply check that the KFAC computation does not modify the `.grad` attribute of the model parameters.

That was a good idea. I found one more line that needed to be removed because we were setting the gradients to zero after KFAC's backward passes.

f-dangel:
Please take a quick second look.

runame (approved):
Perfect, all good now, will merge.

PR #109, opened 2024-05-01
[FIX] Scaling of empirical/MC Fisher for output with more than two dimensions

### LINKED ISSUE #108: [BUG] Scaling of (KFAC) empirical/MC Fisher broken in some cases when using mean reduction for loss
The `reduction_factor` is [assumed](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/gradient_moments.py#L66-L68) to be just the dataset size in the implementation of the empirical Fisher, which is incorrect in the case of `MSELoss`, `BCEWithLogitsLoss`, and in some cases also for `CrossEntropyLoss` (when the model output has more than two dimensions). KFAC with `fisher_type="empirical"` also requires a change in the scaling, similar to [here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/kfac.py#L578-L587).

The same applies to `FisherMCLinearOperator`, but not for KFAC with `fisher_type="mc"`.

runame:
Resolves #108.

@f-dangel and me will investigate different definitions for the Fisher in these cases, but for now we will stick to the here implemented convention.

### DIFF
--- a/test/cases.py
+++ b/test/cases.py
@@ -3,6 +3,7 @@
 from collections import UserDict
 from collections.abc import MutableMapping
 from test.utils import (
+    WeightShareModel,
     binary_classification_targets,
     classification_targets,
     get_available_devices,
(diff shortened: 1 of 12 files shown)


test/cases.py
```diff
@@ -38,28 +39,28 @@ def forward(self, data: MutableMapping):
...
     # same as above, but uses reduction='sum'
     {
-        "model_func": lambda: Sequential(Linear(10, 5), ReLU(), Linear(5, 2)),
+        "model_func": lambda: Sequential(Linear(10, 5), ReLU(), Linear(5, 3)),
         "loss_func": lambda: CrossEntropyLoss(reduction="sum"),
         "data": lambda: [
-            (rand(3, 10), classification_targets((3,), 2)),
-            (rand(4, 10), classification_targets((4,), 2)),
+            (rand(3, 10), classification_targets((3,), 3)),
+            (rand(4, 10), classification_targets((4,), 3)),
```
f-dangel, line 63: Any reason why these cases changed?

runame: Not really, I think I simply preferred having a test case with more than 2 classes, but didn't have a particular failure mode in mind.

PR #110, opened 2024-05-03
[FIX] KFAC scale for `loss_average="batch+sequence"`

### LINKED ISSUE #107: [BUG] Scaling of gradient covariance Kronecker factor in KFAC for `loss_average="batch+sequence"` can be off
When the shape of the gradient of the loss w.r.t. a layer output has shape `(N, R, D_out)`, the model output has shape `(N, R', C)` (assuming MSE loss, otherwise `(N, C, R)`) with `R != R'`, and `loss_average="batch+sequence"`, then the correction of the scale of the gradient covariance Kronecker factor in KFAC ([here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/kfac.py#L730-L731)) is off: `sequence_length` (also part of `num_loss_terms`) is `R`, but should be `R'`.

runame:
Resolves #107.

Please review and merge #109 first since this PR is based on that branch.

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -122,6 +122,7 @@ class KFACLinearOperator(_LinearOperator):
         mc_samples: int = 1,
         kfac_approx: str = "expand",
         loss_average: Union[None, str] = "batch",
+        num_per_example_loss_terms: Optional[int] = None,
         separate_weight_and_bias: bool = True,
         num_data: Optional[int] = None,
         batch_size_fn: Optional[Callable[[MutableMapping], int]] = None,
@@ -188,6 +189,9 @@ class KFACLinearOperator(_LinearOperator):
                 language modeling. If ``None``, the loss function is a sum. This
                 argument is used to ensure that the preconditioner is scaled
                 consistently with the loss and the gradient. Default: ``"batch"``.
+            num_per_example_loss_terms: Number of per-example loss terms. If ``None``,
+                it is inferred from the data at the cost of one traversal through
+                the data loader. Defaults to ``None``.
             separate_weight_and_bias: Whether to treat weights and biases separately.
                 Defaults to ``True``.
             num_data: Number of data points. If ``None``, it is inferred from the data
@@ -205,6 +209,8 @@ class KFACLinearOperator(_LinearOperator):
                 reduction is ``'sum'``.
             ValueError: If ``fisher_type != 'mc'`` and ``mc_samples != 1``.
             ValueError: If ``X`` is not a tensor and ``batch_size_fn`` is not specified.
+            ValueError: If the number of loss terms is not divisible by the number of
+                data points.
         """
         if not isinstance(loss_func, self._SUPPORTED_LOSSES):
             raise ValueError(
@@ -252,14 +258,6 @@ class KFACLinearOperator(_LinearOperator):
         self._gradient_covariances: Dict[str, Tensor] = {}
         self._mapping = self.compute_parameter_mapping(params, model_func)
 
-        # Determine the number of per-example loss terms
-        _, y = next(iter(data))
-        # Assumes that the batch dimension is the first dimension of target `y`
-        if isinstance(loss_func, CrossEntropyLoss):
-            self._num_per_example_loss_terms = y[1:].numel()
-        else:
-            self._num_per_example_loss_terms = y.shape[1:-1].numel()
-
         # Properties of the full matrix KFAC approximation are initialized to `None`
         self._reset_matrix_properties()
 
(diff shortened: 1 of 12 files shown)


curvlinops/kfac.py
```diff
@@ -188,6 +189,9 @@ def __init__(
                 language modeling. If ``None``, the loss function is a sum. This
                 argument is used to ensure that the preconditioner is scaled
                 consistently with the loss and the gradient. Default: ``"batch"``.
+            num_per_example_loss_terms: Number of per-example loss terms. If ``None``,
+                it is inferred from the data at the cost of one traversal through
```
f-dangel, line 193: Can we add a complementary explanation how this number is related to the dimensionality of the random variable whose Fisher we're KFAC-approximating?

curvlinops/kfac.py
```diff
@@ -261,12 +266,57 @@ def __init__(
...
+            # Determine the number of per-example loss terms
+            num_loss_terms = sum(
+                (
+                    y.numel()
+                    if isinstance(self._loss_func, CrossEntropyLoss)
+                    else y.shape[:-1].numel()
+                )
+                for (_, y) in self._loop_over_data(desc="_num_per_example_loss_terms")
+            )
+            if num_loss_terms % self._N_data != 0:
```
f-dangel, line 311: Q: Should we `assert` here that all `y`s are of identical shape except for the batch dimension?

curvlinops/kfac.py
```diff
@@ -261,12 +266,57 @@ def __init__(
...
+                raise ValueError(
+                    "The number of loss terms must be divisible by the number of data "
+                    f"points; num_loss_terms={num_loss_terms}, N_data={self._N_data}."
+                )
+            self._num_per_example_loss_terms = int(num_loss_terms / self._N_data)
```
f-dangel, line 316: Could use `num_loss_terms // self._N_data`

test/test_kfac.py
```diff
@@ -436,17 +440,26 @@ def test_multi_dim_output(
...
     if isinstance(loss_func, MSELoss):
         data = [
-            (rand(2, 7, 5, 5), regression_targets((2, 7, 5, 3))),
-            (rand(4, 7, 5, 5), regression_targets((4, 7, 5, 3))),
+            (X1, regression_targets((2, 7, 5, 3))),
+            (X2, regression_targets((4, 7, 5, 3))),
+        ]
+        manual_seed(711)
+        model = Sequential(Linear(5, 4), Linear(4, 3)).to(dev)
+    elif issubclass(loss, BCEWithLogitsLoss):
```
f-dangel, line 452: Is this `issubclass` intentional, or should this be `isinstance`?

test/test_kfac.py
```diff
@@ -497,6 +510,83 @@ def test_multi_dim_output(
...
+        params,
+        data,
+        fisher_type=fisher_type,
+        loss_average=None,
+    )
+    # FOOF does not scale the gradient covariances, even when using a mean reduction
+    if fisher_type != "forward-only":
+        # Simulate a mean reduction by manually scaling the gradient covariances
+        loss_term_factor = 32 * 32  # number of spatial locations of model output
+        if issubclass(loss, (MSELoss, BCEWithLogitsLoss)):
```
f-dangel, line 569: Should this be `isinstance`?

test/test_kfac.py
```diff
@@ -497,6 +510,83 @@ def test_multi_dim_output(
...
+            loss_term_factor *= 3  # MSE loss averages over number of output channels
```
f-dangel, line 570: Can we assign `3` to a variable with a meaningful name?

f-dangel (commented):
Some questions; might need a second round

curvlinops/kfac.py
runame: The reason I don't do this is that then I have to assume that the batch dimension is the first one  (or some other one) -- maybe not a bad assumption, but we have a test (`test_multi_dim_output`) that relies on being able to use flattened labels. This is also the main reason I don't just determine `num_per_example_loss_terms` from the targets of a single batch.

Not really sure right now what is best here, so let me know if you have a preferred solution.

test/test_kfac.py
runame: It is intentional, since the class is passed as an argument and not a class instance.

test/test_kfac.py
runame: Same as above.

curvlinops/kfac.py
runame: Please check if this is sufficient.

curvlinops/kfac.py
f-dangel: I don't have a better solution at the moment.

f-dangel (approved):

PR #112, opened 2024-05-10
[FIX] Scaling of MC Fisher for `MSELoss` and `BCEWithLogitsLoss` with mean reduction

### LINKED ISSUE #111: [BUG] `FisherMCLinearOperator` is off by a factor of `C` when using `MSELoss` or `BCEWithLogitsLoss` with mean reduction
We scale the loss by `1/C` [here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/fisher.py#L231-L235), but also implicitly [here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/fisher.py#L293) for `MSELoss` and [here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/fisher.py#L327) for `BCEWithLogitsLoss`. This was not caught by tests since the expensive tests in `test_fisher.py` are skipped for the GitHub workflow.

runame:
Fixes #111.

### DIFF
--- a/curvlinops/fisher.py
+++ b/curvlinops/fisher.py
@@ -225,7 +225,7 @@ class FisherMCLinearOperator(_LinearOperator):
 
         grad_output = self.sample_grad_output(output, self._mc_samples, y)
 
-        # Adjust the scale depending on the loss reduction used
+        # Adjust the scale depending on the loss function and reduction used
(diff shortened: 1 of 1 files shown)


f-dangel (approved):
LGTM

PR #117, opened 2024-05-29
Remove `loss_average` argument of `KFACLinearOperator`

runame:
Since we added the `num_per_example_loss_terms` recently, we could remove the `loss_average` argument. I think there is no use case where it adds any additional benefit, and removing it will simplify the user experience significantly and it avoids user errors -- when the user doesn't pass any value for `num_per_example_loss_terms` it will still lead to the correct result, there will just be the overhead of one dataset pass.

Moreover, I created `Enums` for the string valued arguments in `KFACLinearOperator` (`FisherType` and `KFACType`) and added a custom `TypeVar` for the input/ouput of `torch_matmat`/`torch_matvec` to support static type checkers (this was a request by @BrunoKM).

### DIFF
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -1,7 +1,7 @@
 """Implements linear operator inverses."""
 
 from math import sqrt
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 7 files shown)


f-dangel (approved):
LGTM!

PR #119, opened 2024-06-12
[FIX] Checks if `Enum` contains string

### LINKED ISSUE #118: [BUG] Check if string is in `Enum` incompatible with older Python versions
Check if string is in `Enum` is incompatible with older Python versions, e.g., [this](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/kfac.py#L224) will raise an error.

runame:
Fixes #118.

Also, I have added the `FisherType` and `KFACType` Enums to `__init__` to make them easy to import.

### DIFF
--- a/curvlinops/__init__.py
+++ b/curvlinops/__init__.py
@@ -12,7 +12,7 @@ from curvlinops.inverse import (
     NeumannInverseLinearOperator,
 )
 from curvlinops.jacobian import JacobianLinearOperator, TransposedJacobianLinearOperator
-from curvlinops.kfac import KFACLinearOperator
+from curvlinops.kfac import FisherType, KFACLinearOperator, KFACType
 from curvlinops.norm.hutchinson import HutchinsonSquaredFrobeniusNormEstimator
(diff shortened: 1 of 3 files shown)


wiseodd:
Seems like the issue is actually because of a newer version of Python. Your check works in Python 3.8, but fails in Python 3.9 (the default of `laplace-torch`). 

This works in Python 3.9:
```python
In [5]: [v.value for  v in FisherType]
Out[5]: ['type-2', 'mc', 'empirical', 'forward-only']
```

wiseodd:
PS. Python 3.8 will be deprecated soon https://devguide.python.org/versions/.

runame:
@wiseodd Have you tried the fix in this PR? I also tested it with Python 3.11 and it worked.

wiseodd:
Yes, the fix here works. 

Can't wait to finally free from the shackles of Python < 3.10. Pattern matching, [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum), Rust/cargo-like compiler messages... 🤤

PR #120, opened 2024-07-02
[REF] Improve numerical stability of Frobenius estimator

f-dangel:
Instead of estimating the Frobenius norm by `v.T @ (A.T @ A @ v)`, use `w.T @ w` where `w = A @ v`. This 
1. halves the number of `matvec`s
2. improves numerical stability (result cannot be negative), and 
3. does not require the linear operator `A` to define `.T`.


runame (approved):
Very nice, LGTM!

PR #121, opened 2024-07-02
[REF] Deprecate `pkg_resources`, use `importlib` & `packaging`

f-dangel:
Had to explicitly install `packaging` and `wheels` in the CI because otherwise it would crash with an `ImportError` when trying to `import packaging` (note that only installing `packaging` was not sufficient).


runame (approved):
LGTM

PR #125, opened 2024-08-07
[FIX] Output shapes of KFAC's `matmat`/`matvec`

### LINKED ISSUE #124: `A.torch_matvec()` changes the shape of tensors
When `A` is a `KFACLinearOperator` or `KFACInverseLinearOperator` for model parameters given by a list of tensors, the expected behaviour when applying `A` to a list of tensors of the same shape would be that it returns a list of the same shape. This is not the case:
```python
params = [torch.zeros(5,4), torch.zeros(5, 3, 3, 3)]
A = KFACLinearOperator(
    ...
    params,
    ...
)
v = [torch.randn_like(p) for p in params]
y = A.torch_matvec(v)
print([y_elem.shape for y in y_elem])
# Expected output: same shapes as `v` and `params`:
>>> [(5, 4), (5, 3, 3, 3)]
# Actual output: flattened to a 2D tensor:
>>> [(5, 4), (5, 27)]
```

runame:
@BrunoKM Fixes #124.

### DIFF
--- a/curvlinops/inverse.py
+++ b/curvlinops/inverse.py
@@ -627,6 +627,8 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
         for mod_name, param_pos in self._A._mapping.items():
             # retrieve the inverses of the Kronecker factors from cache or invert them
             aaT_inv, ggT_inv = self._compute_or_get_cached_inverse(mod_name)
+            # cache the weight shape to ensure correct shapes are returned
+            weight_shape = M_torch[param_pos["weight"]].shape
 
             # bias and weights are treated jointly
             if (
@@ -644,6 +646,11 @@ class KFACInverseLinearOperator(_InverseLinearOperator):
                 M_torch = self._separate_left_and_right_multiply(
                     M_torch, param_pos, aaT_inv, ggT_inv
                 )
+
+            # restore original shapes
+            M_torch[param_pos["weight"]] = M_torch[param_pos["weight"]].view(
+                weight_shape
+            )
 
         if return_tensor:
             M_torch = cat([rearrange(M, "k ... -> (...) k") for M in M_torch])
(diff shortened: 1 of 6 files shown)


curvlinops/inverse.py
```diff
@@ -645,6 +648,12 @@ def torch_matmat(self, M_torch: ParameterMatrixType) -> ParameterMatrixType:
                     M_torch, param_pos, aaT_inv, ggT_inv
                 )
 
+            # restore original shapes
+            if "weight" in param_pos:
```
BrunoKM, line 652: Is there ever a case when a bias might be >2d (e.g. a CNN on 3D data or similar)? Or would a bias not be reshaped, so restoring the shape for biases is not an issue?

curvlinops/inverse.py
```diff
@@ -627,6 +627,9 @@ def torch_matmat(self, M_torch: ParameterMatrixType) -> ParameterMatrixType:
         for mod_name, param_pos in self._A._mapping.items():
             # retrieve the inverses of the Kronecker factors from cache or invert them
             aaT_inv, ggT_inv = self._compute_or_get_cached_inverse(mod_name)
+            # cache the weight shape to ensure correct shapes are returned
+            if "weight" in param_pos:
+                weight_shape = M_torch[param_pos["weight"]].shape
```
f-dangel, line 632: This is a stylistic comment: I would prefer to use `params` to reshape rather than the matrix supplied by the user, because this will reduce bugs due to user-supplied data:
```
# infer number of vectors onto which KFAC is applied
(V,) = {M.shape[0] for M in M_torch}
...
# look up parameter shapes of current block
w, b = self._params[w_pos], self._params[b_pos]
...
# explicitly reshape to un-flatten flattened dimensions when applying the Kronecker product
M_torch[w_pos] = (...).reshape(V, *w.shape)
M_torch[b_pos] = (...).reshape(V, *b.shape)
```
What do you think?

runame: We do explicitly check the shapes of the user supplied data [here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/kfac.py#L383-L401), so this shouldn't make a difference?

f-dangel (approved):

PR #128, opened 2024-09-18
[REQ] Deprecate python 3.8

f-dangel:
Python 3.8 will reach its end of life in October 2024.
Therefore, we will deprecate it and require at least Python 3.9 in future releases.

### DIFF
--- a/.conda_env.yml
+++ b/.conda_env.yml
@@ -3,7 +3,7 @@ channels:
   - pytorch
   - defaults
 dependencies:
-  - pip=21.2.4
-  - python=3.8.5
+  - python=3.9.16
+  - pip=23.1.2
   - pip:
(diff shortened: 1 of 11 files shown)


.conda_env.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - pytorch
   - defaults
 dependencies:
-  - pip=21.2.4
-  - python=3.8.5
+  - python=3.9.16
```
runame, line 6: Any reason for `3.9.16` specifically? Latest is `3.9.20` I think.

runame (approved):
LGTM

.conda_env.yml
f-dangel: The only reason is that I know this `python`+`pip` versions are compatible with each other.

PR #130, opened 2024-09-21
[ADD] Minimal linear operator interface for PyTorch

f-dangel:
Long-term, I want to add native PyTorch support for linear operators in `curvlinops` to address inefficiencies like #71, but also to clearly separate PyTorch from SciPy so that it will be easier to tackle features like supporting distributed settings. 

This PR is a first step towards this goal.

From an API perspective, I plan to keep the constructor of all existing linear operators identical. The only backward-incompatible change will be that the produced linear operator will be purely PyTorch. To obtain the old behaviour one has to call `.to_scipy()` after the constructor.

**Old:** `H = HessianLinearOperator(...)`
**Planned new:** `H = HessianLinearOperator(...).to_scipy(dtype=...)`

The PR defines a linear operator interface in PyTorch which allows easy export to SciPy linear operators.
Importantly, the interface can multiply onto vectors/matrices represented by single `Tensor`s, or a `List[Tensor]`, which is more common in PyTorch. It verifies the input and output formats and all methods that need to be implemented assume the (more natural) tensor list format.

The next steps will be:
- Define a base class `CurvatureLinearOperator` that replicates `curvlinops._base._LinearOperator` but inherits from our `PyTorchLinearOperator`, rather than `scipy.sparse.linalg.LinearOperator`.
- Migrate each supported linear operator to inherit from `CurvatureLinearOperator`. I already tried that for the Hessian and was able to migrate without breaking the tests. I will set up a separate PR to keep the diffs manageable
- Once all operators have been migrated (and probably we can get rid of a lot of boilerplate to check shapes, e.g. in KFAC), we can remove the current base class in `curvlinops._base`.

Let me know if this makes sense.

### DIFF
--- a/test/test__torch_base.py
+++ b/test/test__torch_base.py
@@ -39,44 +39,44 @@ def test_output_formatting():
     in_shape = [(2, 3), (4, 5)]
     out_shape = [(2, 3), (4, 6)]  # NOTE that this will trigger an error
 
-    I = IdentityLinearOperator(in_shape, out_shape)
-    assert I._in_shape_flat == [6, 20]
-    assert I._out_shape_flat == [6, 24]
-    assert I.shape == (30, 26)
+    Id = IdentityLinearOperator(in_shape, out_shape)
+    assert Id._in_shape_flat == [6, 20]
+    assert Id._out_shape_flat == [6, 24]
+    assert Id.shape == (30, 26)
 
     # using valid input vectors/matrices will trigger errors because we
     # initialized the identity with different input/output spaces
     with raises(ValueError):
-        _ = I @ [zeros(2, 3), zeros(4, 5)]  # valid vector in list format
+        _ = Id @ [zeros(2, 3), zeros(4, 5)]  # valid vector in list format
 
     with raises(ValueError):
-        _ = I @ [zeros(2, 3, 6), zeros(4, 5, 6)]  # valid matrix in list format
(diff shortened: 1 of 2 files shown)


test/test__torch_base.py
```diff
@@ -0,0 +1,82 @@
...
+
+    X = [zeros(2, 3), zeros(4, 5)]  # vector in tensor list format
+    IdX = Id @ X
+    assert len(IdX) == len(X) and all(Idx.allclose(x) for Idx, x in zip(IdX, X))
+
+    X = [zeros(2, 3, 6), zeros(4, 5, 6)]  # matrix in tensor list format
+    IdX = Id @ X
+    assert len(IdX) == len(X) and all(Idx.allclose(x) for Idx, x in zip(IdX, X))
+
+    X = zeros(26)  # vector in tensor format
```
runame, line 76: Does it make sense to test multiplication with 0dim and 1dim vector?

runame (approved):
I haven't thought much about what functions have to implemented but this LGTM! This will definitely be a useful refactor and enable/simplify new functionality.

test/test__torch_base.py
f-dangel: Could you elaborate what you mean by 0-dim vector? 
Re 1-dim vectors: isn't `X` here a one-dimensional vector?

f-dangel: I will go ahead and merge this. Let me know if you have thoughts on testing the 0-dim case.

PR #131, opened 2024-09-23
[CI | FIX] Improve and run Fisher-MC tests on `main`

f-dangel:
I noticed that the Monte-Carlo tests were not running anymore because we switched from `master` to `main`.
This PR re-activates them, slightly refactors them to make them faster, and adapts their tolerances to account for cases we added but never tested.


runame (approved):
LGTM

PR #133, opened 2024-10-02
[BUG] Wrong dtype for random vectors used in KFAC's deterministic checks

### LINKED ISSUE #132: `KFACLinearOperator` deterministic checks break when using double precision
I want to use my model in double precision, so I need to pass in data in double precision, but I will have error, here is I reproduced the error with toy example by changing precision:
```Python
# make deterministic
torch.manual_seed(0)
numpy.random.seed(0)

N = 4
D_in = 7
D_hidden = 5
D_out = 3
DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")

X = torch.rand(N, D_in, dtype=torch.float64).to(DEVICE)
y = torch.rand(N, D_out, dtype=torch.float64).to(DEVICE)

model = nn.Sequential(
    nn.Linear(D_in, D_hidden),
    nn.ReLU(),
    nn.Linear(D_hidden, D_hidden),
    nn.Sigmoid(),
    nn.Linear(D_hidden, D_out),
).double().to(DEVICE)
params = [p for p in model.parameters() if p.requires_grad]
data = [(X, y)]
loss_function = nn.MSELoss(reduction="mean").to(DEVICE)
KH = KFACLinearOperator(model, loss_function, params, data)
```

And the error I got is:

> Traceback (most recent call last):
>   File "/home/d7liao/TGDA-Attack/example.py", line 34, in <module>
>     KH = KFACLinearOperator(model, loss_function, params, data)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/curvlinops/kfac.py", line 284, in __init__
>     raise e
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/curvlinops/kfac.py", line 282, in __init__
>     self._check_deterministic()
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/curvlinops/_base.py", line 195, in _check_deterministic
>     mat_v1 = self @ v
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/scipy/sparse/linalg/_interface.py", line 455, in __matmul__
>     return self.__mul__(other)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/scipy/sparse/linalg/_interface.py", line 411, in __mul__
>     return self.dot(x)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/scipy/sparse/linalg/_interface.py", line 444, in dot
>     return self.matvec(x)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/scipy/sparse/linalg/_interface.py", line 236, in matvec
>     y = self._matvec(x)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/scipy/sparse/linalg/_interface.py", line 203, in _matvec
>     return self.matmat(x.reshape(-1, 1))
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/scipy/sparse/linalg/_interface.py", line 340, in matmat
>     Y = self._matmat(X)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/curvlinops/kfac.py", line 527, in _matmat
>     M_torch = self.torch_matmat(M_torch)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/curvlinops/kfac.py", line 463, in torch_matmat   
>     M_torch[pos] = einsum(
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/einops/einops.py", line 907, in einsum
>     return get_backend(tensors[0]).einsum(pattern, *tensors)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/einops/_backends.py", line 287, in einsum
>     return self.torch.einsum(pattern, *x)
>   File "/home/d7liao/anaconda3/envs/traits/lib/python3.10/site-packages/torch/functional.py", line 378, in einsum
>     return _VF.einsum(equation, operands)  # type: ignore[attr-defined]
> RuntimeError: expected scalar type Float but found Double

> 
Thanks in advance

f-dangel:
Resolves #132.

The problem was that the deterministic checks generate random numpy vectors which were then converted to the incorrect PyTorch data dtype.

Interestingly, I found that the other linear operators, such as Hessian, GGN, Fisher, EF, were not affected by this problem. I'm not 100% sure why. @runame any thoughts?


curvlinops/_base.py
```diff
@@ -324,7 +325,11 @@ def _postprocess(self, M_list: List[Tensor]) -> ndarray:
             concatenated dimensions over all list entries.
         """
         result = [rearrange(M, "k ... -> (...) k") for M in M_list]
-        return cat(result, dim=0).cpu().numpy()
+        result = cat(result)
+        # calling .numpy() on a BF-16 tensor is not supported, see
+        # (https://github.com/pytorch/pytorch/issues/90574)
+        result = result.float() if result.dtype == bfloat16 else result
+        return result.cpu().numpy().astype(self.dtype)
```
runame, line 332: Will `self.dtype` ever be anything else but `float32`? It seems it is hard-coded [here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/_base.py#L102), but not sure if this is overridden somewhere else.

runame (commented):
I'm a bit confused, haven't looked at how dtypes are handled in a while. The core difference between KFAC and the other linops is that in KFAC we first calculate state (the Kronecker factors), which will be stored in the dtype of the model/data, whereas in the other linops there is no state and the outputs of the forward/backward pass are only interacting with the vector (in the MVP) via BackPACK's `hessian_vector_product` or `ggn_vector_product_from_plist`.

Assuming that `self.dtype` is always `float32` (see my question below), it makes sense that KFAC fails, since its state will be `float64` and the vector will be `float32`. However, I checked the BackPACK functions mentioned above and don't see any dtype conversions, so not sure why don't fail as well?

curvlinops/_base.py
f-dangel: It is hard-coded. A better way to resolve this would be to expose the `dtype` in the constructor of all linear operators, but I'm hesitant to do this as I'm planning to deprecate the base class soon.

curvlinops/_base.py
```diff
@@ -302,7 +303,7 @@ def _preprocess(self, M: ndarray) -> List[Tensor]:
             M = M.astype(self.dtype)
         num_vectors = M.shape[1]
 
-        result = from_numpy(M).to(self._device)
+        result = from_numpy(M).to(self._device, self._torch_dtype)
```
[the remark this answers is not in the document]
f-dangel: Thanks, changed that.

curvlinops/_base.py
f-dangel: I think we can ignore for now that the dtype issue does not present itself for the state-less linear operators.

runame (approved):
Let's just go ahead with this, I suspect the dtype logic will change anyway with the coming refactors.

PR #134, opened 2024-10-06
[DOC] Add more matrices to visual tour, prettify plots with `tueplots`

f-dangel:
Adds the block-diagonal Hessian, MC-Fisher, and KFAC to the visual tour tutorial.
Prettifies the visualizations using `tueplots`.

PR #139, opened 2024-10-21
[DOC] Prettify plots in spectral density tutorial

f-dangel:
Please review #134 first.

Uses `tueplots` to make the plots in the spectral density tutorial prettier.


f-dangel:
@runame ready for review. Just a quick pass should be enough, it is mostly plotting code.

PR #140, opened 2024-10-21
[DOC] Add comparison with power iteration in eigenvalue example

f-dangel:


docs/examples/basic_usage/example_eigenvalues.py
```diff
@@ -104,3 +108,128 @@
...
+
+    return eigenvalues, eigenvectors
+
+
+# %%
+#
+# Let's compute the top-3 eigenvalues via power iteration and verify they roughly match.
+# Note that we are using a smaller :code:`tol` value than the PyHessian default value
+# here to get better convergence, and we have to use relatively large tolerances for the
+# comparison (which we didn't when comparing :code:`eigsh` with :code:`eigh`).
```
werkaaa, line 192: Suggestion: It feels like, the subsentence in the brackets lacks a verb like "do": "...which we didn't do when comparing...".

docs/examples/basic_usage/example_eigenvalues.py
```diff
@@ -104,3 +108,128 @@
...
+# This indicates that the power method achieves poorer accuracy than :code:`eigsh`. But
+# does it therefore require fewer matrix-vector products? To answer this, let's turn on
+# the linear operator's progress bar, which allows us to count the number of
+# matrix-vector products invoked by both eigen-solvers:
+
+H = HessianLinearOperator(model, loss_function, params, data, progressbar=True)
+
+# determine number of matrix-vector products used by `eigsh`
+with StringIO() as buf, redirect_stderr(buf):
+    top_k_evals, _ = scipy.sparse.linalg.eigsh(H, k=k, which="LA")
```
werkaaa, line 209: The power method will give us the eigenvalue of the largest magnitude, right? If so, for a fair comparison, I would also tell `eigsh` to do the same and set `which='LM'`.

I have the same concern for line 196: aren't we now comparing the largest magnitude eigenvalue with the largest algebraic one?

f-dangel: You are right, for the Hessian it seems that the largest magnitudes are at the same time the largest algebraic. I've corrected it everywhere in the example. Thanks for pointing out this mistake.

PR #142, opened 2024-10-25
[REF] Replicate base class, but inherit from PyTorch linear operator

f-dangel:
This PR introduces a class `CurvatureLinearOperator`, which is almost identical to the current base class `_LinearOperator`, but inherits from `PyTorchLinearOperator`.

The next steps will be to (i) migrate each linear operator to inherit from `CurvatureLinearOperator`, rather than `_LinearOperator`, (ii) delete `_LinearOperator`.

### DIFF
--- a/curvlinops/_torch_base.py
+++ b/curvlinops/_torch_base.py
@@ -128,6 +128,9 @@ class PyTorchLinearOperator:
                 This is useful for post-processing the multiplication's result.
             is_vec: Whether the input is a vector or a matrix.
             num_vecs: The number of vectors represented by the input.
+
+        Raises:
+            ValueError: If the input format is invalid.
         """
         if isinstance(X, Tensor):
             list_format = False
@@ -295,10 +298,10 @@ class PyTorchLinearOperator:
     def _infer_device(self) -> device:
         """Infer the linear operator's device.
 
-        Returns:
+        Returns:  # noqa: D402
             The device of the linear operator.
 
-        Raises: # noqa: D402
+        Raises:
             NotImplementedError: Must be implemented by subclasses.
         """
         raise NotImplementedError
@@ -306,10 +309,10 @@ class PyTorchLinearOperator:
     def _infer_dtype(self) -> dtype:
         """Infer the linear operator's data type.
 
-        Returns:
+        Returns: # noqa: D402
             The data type of the linear operator.
 
-        Raises: # noqa: D402
+        Raises:
             NotImplementedError: Must be implemented by subclasses.
         """
         raise NotImplementedError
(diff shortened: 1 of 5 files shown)


curvlinops/_torch_base.py
```diff
@@ -347,3 +352,344 @@ def f_scipy(X: numpy.ndarray) -> numpy.ndarray:
...
+            ValueError: If any block size is not positive.
+            ValueError: If ``X`` is not a tensor and ``batch_size_fn`` is not specified.
+        """
+        if isinstance(next(iter(data))[0], MutableMapping) and batch_size_fn is None:
+            raise ValueError(
+                "When using dict-like custom data, `batch_size_fn` is required."
+            )
+
+        in_shape = [tuple(p.shape) for p in params] if in_shape is None else in_shape
+        out_shape = [tuple(p.shape) for p in params] if in_shape is None else in_shape
```
f-dangel, line 442: Typo: should be `out_shape`

curvlinops/_torch_base.py
```diff
@@ -347,3 +352,344 @@ def f_scipy(X: numpy.ndarray) -> numpy.ndarray:
...
+        Args:
+            desc: Description for the progress bar. Will be ignored if progressbar is
+                disabled.
+            add_device_to_desc: Whether to add the device to the description.
+                Default: ``True``.
+
+        Yields:
+            Mini-batches ``(X, y)``.
+        """
+        data_iter = iter(self._data)
```
f-dangel, line 541: Remove `iter`

curvlinops/_torch_base.py
```diff
@@ -347,3 +352,344 @@ def f_scipy(X: numpy.ndarray) -> numpy.ndarray:
...
+        """Apply the mini-batch matrix to a vector.
+
+        Args:
+            X: Input to the DNN.
+            y: Ground truth.
+            M: Matrix in list format (same shape as trainable model parameters with
+                additional trailing dimension of size number of columns).
+
+        Returns: # noqa: D402
+           Result of matrix-multiplication in list format.
```
runame, line 520: This seems to be inconsistent with the type hint `Tuple[Tensor]`.

runame (approved):
LGTM!

Would be good to add unit tests, at least of all public methods.

curvlinops/_torch_base.py
f-dangel: Fixed.

f-dangel:
Added the following test: In this PR, `CurvatureLinearOperator` has two public functions: `to_device`, and `gradient_and_loss`. The `to_device` function will be deprecated soon, therefore I am not adding a test here (it is also called by the other linear operators in their deterministic checks, therefore tested indirectly). I added an explicit test which makes sure that the loss and gradient are correctly accumulated over a data set. This functionality was tested indirectly before by making sure the curvature matrix-vector products are correctly accumulated over a data set.

PR #146, opened 2024-10-27
[ADD] Change GGN to purely PyTorch

f-dangel:
Only review after #142, #145.

### DIFF
--- a/curvlinops/ggn.py
+++ b/curvlinops/ggn.py
@@ -8,11 +8,11 @@ from typing import List, Tuple, Union
 from backpack.hessianfree.ggnvp import ggn_vector_product_from_plist
 from torch import Tensor, zeros_like
 
-from curvlinops._base import _LinearOperator
+from curvlinops._torch_base import CurvatureLinearOperator
 
 
-class GGNLinearOperator(_LinearOperator):
-    r"""GGN as SciPy linear operator.
+class GGNLinearOperator(CurvatureLinearOperator):
+    r"""Linear operator for the generalized Gauss-Newton matrix of an empirical risk.
 
     Consider the empirical risk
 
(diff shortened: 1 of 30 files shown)


curvlinops/ggn.py
```diff
@@ -39,47 +39,42 @@ class GGNLinearOperator(_LinearOperator):
...
+            M: Matrix to be multiplied with in tensor list format.
                 Tensors have same shape as trainable model parameters, and an
-                additional leading axis for the matrix columns.
+                additional trailing axis for the matrix columns.
 
         Returns:
             Result of GGN multiplication in list format. Has the same shape as
-            ``M_list``, i.e. each tensor in the list has the shape of a parameter and a
-            leading dimension of matrix columns.
+            ``M_``, i.e. each tensor in the list has the shape of a parameter and a
```
runame, line 63: `M` instead of `M_`?

docs/examples/basic_usage/example_fisher_monte_carlo.py
```diff
@@ -78,7 +78,7 @@
 # Fisher and compute their matrix representations by multiplying them onto the
 # identity matrix:
 
-GGN = GGNLinearOperator(model, loss_function, params, data)
+GGN = GGNLinearOperator(model, loss_function, params, data).to_scipy()
```
runame, line 81: What is the reason to call `.scipy()` for `GGN` but not for `F`?

runame (approved):
LGTM!

curvlinops/ggn.py
f-dangel: Good catch.

docs/examples/basic_usage/example_fisher_monte_carlo.py
f-dangel: This PR only changes the GGN linear operator to pure PyTorch. The Fisher linear operator is still a normal SciPy linear operator, therefore it does not need to be converted explicitly.

runame: ahh of course, thanks!

PR #148, opened 2024-10-27
[ADD] Change Fisher to purely PyTorch

f-dangel:
Review after #142, #145, #146, #147.


runame (approved):
LGTM!

One of the examples is still broken though, I think you have to call `.to_scipy()` [here](https://github.com/f-dangel/curvlinops/blob/main/docs/examples/basic_usage/example_visual_tour.py#L100).

PR #149, opened 2024-10-27
[ADD] Change KFAC and KFAC inverse to purely PyTorch

f-dangel:
Review after #142, #145, #146, #147, #148, #149.

PR #156, opened 2024-11-29
[FIX] in-place damping bug in `KFACInverseLinearOperator`

runame:
When `use_exact_damping=True` and `KFACLinearOperator._separate_weight_and_bias=True`, the damping will be repeatedly added to the eigenvalues corresponding to the bias parameters with every matrix-vector product.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed/Removed
 
+- Bug in `KFACInverseLinearOperator` where the damping was repeatedly (for every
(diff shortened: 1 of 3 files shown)


f-dangel:
Great catch! Can we merge this after #149? This PR refactors the `KFAC` class a lot, and I'm afraid it might lead to having to resolve a lot of nasty merge conflicts.

runame:
Sure, but this PR only changes a single underscore in `inverse.py`, so should not result in any significant merge conflict. The other changed lines are in the test file.

f-dangel (approved):
LGTM. Could you update the changelog before merging?

PR #158, opened 2024-12-01
[ADD] more test cases for inverse KFAC and fix bug in `KFAC.load_state_dict`

runame:
Follow up to #156, sorry, missed a test case.

_Edit: Added even more test cases and discovered and fixed a bug in KFAC's `load_state_dict` function._

### DIFF
--- a/curvlinops/kfac.py
+++ b/curvlinops/kfac.py
@@ -1147,9 +1147,8 @@ class KFACLinearOperator(_LinearOperator):
             input_covariances_keys = set(self._input_covariances.keys())
             gradient_covariances_keys = set(self._gradient_covariances.keys())
             mapping_keys = set(self._mapping.keys())
-            if (
-                input_covariances_keys != mapping_keys
-                or gradient_covariances_keys != mapping_keys
+            if (input_covariances_keys and input_covariances_keys != mapping_keys) or (
+                gradient_covariances_keys and gradient_covariances_keys != mapping_keys
             ):
                 raise ValueError(
                     "Input or gradient covariance keys in state dict do not match "
--- a/test/test_inverse.py
+++ b/test/test_inverse.py
@@ -754,7 +754,7 @@ def test_KFAC_inverse_save_and_load_state_dict(
     inv_kfac = KFACInverseLinearOperator(
         kfac, damping=1e-2, retry_double_precision=False, cache=cache, **kwargs
     )
-    test_vec = torch.rand(inv_kfac.shape[1])
+    test_vec = random.rand(inv_kfac.shape[1])
     test_mvp = inv_kfac @ test_vec  # triggers inverse computation and maybe caching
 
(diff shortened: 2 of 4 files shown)


curvlinops/kfac.py
```diff
@@ -1147,9 +1147,8 @@ def load_state_dict(self, state_dict: Dict[str, Any]):
             input_covariances_keys = set(self._input_covariances.keys())
             gradient_covariances_keys = set(self._gradient_covariances.keys())
             mapping_keys = set(self._mapping.keys())
-            if (
-                input_covariances_keys != mapping_keys
-                or gradient_covariances_keys != mapping_keys
+            if (input_covariances_keys and input_covariances_keys != mapping_keys) or (
+                gradient_covariances_keys and gradient_covariances_keys != mapping_keys
```
f-dangel, line 1151: Could you explain what this does? I am not familiar with this notation for sets.

f-dangel: I tried understanding what is happening with the following:
```python
a = {1,2}
b = {2,3}

a and b # returns {2,3}
```

test/test_inverse.py
```diff
@@ -704,40 +775,72 @@ def test_KFAC_inverse_save_and_load_state_dict():
...
+        "use_heuristic_damping": use_heuristic_damping,
+    }
+    if use_exact_damping and use_heuristic_damping:
+        return
     inv_kfac = KFACInverseLinearOperator(
-        kfac, damping=1e-2, use_heuristic_damping=True, retry_double_precision=False
+        kfac, damping=1e-2, retry_double_precision=False, cache=cache, **kwargs
     )
+    test_vec = torch.rand(kfac.shape[1])
+    test_mvp = inv_kfac @ test_vec  # triggers inverse computation and maybe caching
```
f-dangel, line 835: Question: Why does `@` work on a PyTorch tensor here? Shouldn't `inv_kfac` process `numpy` arrays, at least before we merge #149?

curvlinops/kfac.py
runame: In Python `!=` has higher precedence than `and`, i.e. each of the expressions in the parentheses here are of type `set and bool`. A set evaluated as a boolean will return `True` if it is non-empty and `False` if it is empty. Hence, this expression is first checking if the set is non-empty and then if it matches the `mapping_keys`. Does this make sense or am I missing something?

test/test_inverse.py
runame: It should work with PyTorch tensors as long as they are on CPU, because of [this](https://github.com/scipy/scipy/blob/v1.14.1/scipy/sparse/linalg/_interface.py#L330-L331). But you're right that this goes against our type hints and only works on CPU, so I'm happy to change to a numpy array here.

f-dangel (approved):
LGTM.

PR #160, opened 2024-12-15
[DEL] Remove `to_device` function, run deterministic tests on original device

f-dangel:
Before this PR, we always run the deterministic checks on CPU.

I thought this made more sense because it does not introduce randomness from GPU computations.
I now think it is better to carry out the deterministic checks on the actual device a linear operator lives on. It will be more meaningful for a user to know that the matrix-vector multiply is non-deterministic on their GPU, or at least somewhat variable.

This PR makes linear operators execute their deterministic checks on the device they are defined on.
As a result, we can deprecate the `to_device` function.

After approval and before merging:
- [x] verify on a GPU environment that all tests pass (I already did that at some point when preparing this PR)


runame (approved):
LGTM.

After [adjusting the numerical tolerances for the tests](https://github.com/f-dangel/curvlinops/pull/160/commits/1eb0b49af9a03d878621f1ad9b36108dc9d6714d), all tests (`make test`) pass on a GPU machine.

PR #162, opened 2024-12-17
[DOC] Demo how to benchmark linear operators

f-dangel:
Adds an example to the documentation that benchmarks run time and memory consumption.

WIP:
- [x] Write content of code cells, walking the user through every step
- [x] Ensure benchmarks work on GPU
- [x] Add larger problems for presentation in the paper


docs/examples/basic_usage/memory_benchmark.py
```diff
@@ -0,0 +1,169 @@
...
+def run_peakmem_benchmark(  # noqa: C901
+    linop_str: str, problem_str: str, device_str: str, op_str: str
+):
+    """Execute the memory benchmark for a given linear operator class and save results.
+
+    Args:
+        linop_str: The linear operator.
+        problem_str: The problem.
+        device_str: The device.
+        write_results: Whether to write the results to a file. Default is ``True``.
```
runame, line 33: This argument doesn't exist and there is not entry for `op_str`.

docs/examples/basic_usage/benchmark_utils.py
```diff
@@ -0,0 +1,146 @@
...
+
+    Args:
+        batch_size: The batch size to use. Default is ``2``.
+
+    Returns:
+        A tuple containing the nanoGPT model, the loss function, and the data.
+    """
+    # download nanogpt_model and import GPT and GPTConfig from it
+    # also download the Shakespeare dataset
+    data_dir = maybe_download_nanogpt_shakespeare()
```
runame, line 95: Is there a reason you use real instead of synthetic data here?

runame (approved):
LGTM

docs/examples/basic_usage/memory_benchmark.py
f-dangel: Thanks

docs/examples/basic_usage/benchmark_utils.py
f-dangel: No. I rather started with the real data set to avoid any issues when adding the GPT model. Will refactor to use synthetic data instead.

PR #164, opened 2024-12-31
[CI] Migrate from `flake8`, `black`, and `isort` to `ruff`

f-dangel:
Only review after #163.


makefile
```diff
@@ -97,10 +97,10 @@ black:
 black-check:
 	@black . --config=black.toml --check
 
-.PHONY: flake8
+.PHONY: ruff
 
-flake8:
-	@flake8 .
+ruff:
+	@ruff check .
```
runame, line 103: Do we not also want to add `--fix`?

pyproject.toml
```diff
@@ -106,3 +99,38 @@ use_parentheses = true
...
+    "W291",
+    # E203 line break before binary operator (replaces W503)
+    "E203",
+    # Line break occurred after a binary operator (replaces W504)
+    "E226",
+    # B905 `zip()` without an explicit `strict=` parameter
+    "B905",
+    # Too many arguments in function definition (9 > 5)
+    "PLR0913",
+]
```
runame, line 129: Should we just add `PLR2004` and `PLW2901` here?

runame (commented):
I think we should also remove the `isort` dependency and add the corresponding `ruff` rules. Also, we could switch to the`ruff` formatter as well, instead of using `black`.

f-dangel:
Migrated formatting and import sorting.

I also tried using the docstring linter of ruff, but it produced too many errors that I don't have time to fix and incorporate in this PR. Might be a useful thing to add in the future though.

PR #166, opened 2025-01-06
[ADD] Implement XTrace trace estimator

f-dangel:


curvlinops/trace/epperly2024xtrace.py
```diff
@@ -0,0 +1,101 @@
...
+    Q, R = qr(A_W)
+    A_Q = A @ Q
+    tr_QT_A_Q = einsum("ij,ji->", Q.T, A_Q)
+
+    # compute the traces in the bases that would result had we left out the i-th
+    # test vector in the QR decomposition
+    RT_inv = inv(R.T)
+    D = 1 / (RT_inv**2).sum(0) ** 0.5
+    S = einsum("ij,j->ij", RT_inv, D)
+    tr_QT_i_A_Q_i = einsum("ij,jk,kl,li->i", S.T, Q.T, A_Q, S)
```
runame, line 62: Maybe more efficient:
```suggestion
    tr_QT_i_A_Q_i = einsum("ij,ki,kl,lj->i", S, Q, A_Q, S)
```
If you stick to the previous version, maybe compute `Q.T` once and reuse it.

runame (approved):
I don't fully understand the algorithm (would have to look at the paper for that), but the implementation LGTM.

curvlinops/trace/epperly2024xtrace.py
f-dangel: Yeah, good idea to use a single einsum. I think the `einsum` string should now `->j`, not `->i`, so I will push the refactor manually. One curiosity: I think it is necessary to set `optimize=True`, because in NumPy the contraction is not optimized by default.

PR #167, opened 2025-01-07
[ADD] Implement XDiag diagonal estimator

f-dangel:


curvlinops/diagonal/epperly2024xtrace.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Compute and average the diagonals in the bases {Q_i} that would result had we left
+    # out the i-th test vector in the QR decomposition. This follows by considering
+    # diag(Q_i QT_i A) and using the relation Q_i QT_i = Q (I - s_i sT_i) QT, where the
+    # s_i are given by:
+    RT_inv = inv(R.T)
+    D = 1 / (RT_inv**2).sum(0) ** 0.5
+    S = einsum("ij,j->ij", RT_inv, D)
+    # Further simplification then leads to
+    diagonal = diag_Q_QT_A - einsum("ij,ji->i", Q @ S, S.T @ QT_A) / num_vecs
```
runame, line 61: Maybe more efficient with a single einsum:
```suggestion
    diagonal = diag_Q_QT_A - einsum("ij,jk,lk,lm->i", Q, S, S, QT_A) / num_vecs
```

runame (approved):
Same comment as for #166 , the implementation LGTM.

Also, maybe it's possible to avoid some code duplication since the two functions are very similar, but I think the current implementation is just fine.

curvlinops/diagonal/epperly2024xtrace.py
f-dangel: Sounds good. I will push the manual fix (this time I really think the `einsum` string is not the right one 😛) and set `optimize='optimal'`, too.

PR #168, opened 2025-01-07
[REF] Make trace/diagonal estimators functional

f-dangel:


curvlinops/diagonal/hutchinson.py
```diff
@@ -1,27 +1,30 @@
...
+from numpy import column_stack, einsum, ndarray
 from scipy.sparse.linalg import LinearOperator
 
 from curvlinops.sampling import random_vector
 
 
-class HutchinsonDiagonalEstimator:
-    r"""Class to perform diagonal estimation with Hutchinson's method.
+def hutchinson_diag(
+    A: LinearOperator, num_matvecs: int, distribution: str = "rademacher"
```
runame, line 10: No strong opinion, but maybe something like `num_samples` instead of `num_matvecs` is a bit more intuitive in the context of a stochastic estimator?

curvlinops/diagonal/hutchinson.py
```diff
@@ -33,54 +36,46 @@ class HutchinsonDiagonalEstimator:
...
+    if len(A.shape) != 2 or A.shape[0] != A.shape[1]:
+        raise ValueError(f"A must be square. Got shape {A.shape}.")
+    dim = A.shape[1]
+    if num_matvecs >= dim:
+        raise ValueError(
+            f"num_matvecs ({num_matvecs}) must be less than A's size ({dim})."
+        )
+    G = column_stack([random_vector(dim, distribution) for _ in range(num_matvecs)])
+    AG = A @ G
+    return einsum("ij,ij->i", G, AG) / num_matvecs
```
runame, line 81: Can be rewritten as a single einsum.

curvlinops/trace/hutchinson.py
```diff
@@ -34,49 +37,44 @@ class HutchinsonTraceEstimator:
...
+    if len(A.shape) != 2 or A.shape[0] != A.shape[1]:
+        raise ValueError(f"A must be square. Got shape {A.shape}.")
+    dim = A.shape[1]
+    if num_matvecs >= dim:
+        raise ValueError(
+            f"num_matvecs ({num_matvecs}) must be less than A's size ({dim})."
+        )
+    G = column_stack([random_vector(dim, distribution) for _ in range(num_matvecs)])
+
+    return einsum("ij,ij", G, A @ G) / num_matvecs
```
runame, line 80: This can be rewritten as a single einsum operation, without first computing `A @ G`.

curvlinops/diagonal/hutchinson.py
```diff
@@ -33,54 +36,46 @@ class HutchinsonDiagonalEstimator:
...
-        v = random_vector(dim, distribution)
-        Av = self._A @ v
-        return v * Av
+    if len(A.shape) != 2 or A.shape[0] != A.shape[1]:
+        raise ValueError(f"A must be square. Got shape {A.shape}.")
+    dim = A.shape[1]
+    if num_matvecs >= dim:
+        raise ValueError(
+            f"num_matvecs ({num_matvecs}) must be less than A's size ({dim})."
+        )
```
runame, line 78: Not necessary, but you could consider factoring these lines out since they appear in at least three functions (and once slightly modified) to reduce code duplication.

runame (approved):
LGTM.

curvlinops/diagonal/hutchinson.py
f-dangel: I thought so too when I first implemented these methods. After reading more into the literature (Hutch++, XTrace), I think number of matrix-vector queries is better, because it applies to all trace estimators. E.g. for Hutch++, number of samples is different from number of matrix-vector products (because two thirds of the matvecs is used for deterministic trace estimation, one third for 'sampled' trace estimation). Using `num_matvecs` also allows these methods to be compared more easily (i.e. using same number of `matvecs`), because users don't need to know how these methods work in detail.

curvlinops/diagonal/hutchinson.py
f-dangel: You are right if `A` is a `Tensor`. However, in general `A` can be a `LinearOperator`. In this case, we must compute `A @ G`, because expanding `A`'s matrix representation would be too costly.

(The same applies to your other comments on single einsums, I believe. Let me know if I missed something here.)

curvlinops/trace/hutchinson.py
f-dangel: See comment above.

curvlinops/trace/meyer2020hutch.py
```diff
@@ -1,152 +1,106 @@
...
-            v = self._Q[:, i]
-            Av = self._A @ v
-            self._tr_QT_A_Q += dot(v, Av)
+    N = num_matvecs // 3
+
+    # compute the orthogonal basis for the subspace spanned by AS, and evaluate the
+    # exact trace using 2/3 of the available matrix-vector products
+    AS = A @ column_stack([random_vector(dim, distribution) for _ in range(N)])
+    Q, _ = qr(AS)
+    tr_QT_A_Q = einsum("ji,ji", Q, A @ Q)
```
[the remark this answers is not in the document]
f-dangel: See comment above.

curvlinops/diagonal/hutchinson.py
f-dangel: Added.

curvlinops/diagonal/hutchinson.py
runame: Ok makes sense!

runame (commented):
LGTM now.

PR #172, opened 2025-02-18
Stop overriding `model_func` state dictionary when loading state dicts of K-FAC or K-FAC inverse

BrunoKM:
Currently, when calling `.load_state_dict(state_dict)` on a `KFACLinearOperator` or `KFACInverseLinearOperator` object initialised with some `model_func`, this will override the state dictionary of the `model_func`. This might lead to unexpected behaviour, wherein:

**1)** _(minor)_ The user suddenly ends up with parameters of their model overwritten:
```python
model = ...
kfac = KFACLinearOperator(model_func=model, ...)
kfac.load_state_dict(torch.load("kfac_state_dict.torch"))
# Parameters of `model` have been silently overwritten
```
**2)** _(major)_ The references that the user was using no longer point to the same parameters:
```python
feature_extractor = ConvNet(...)  # A PyTorch module
model = nn.Sequential(
    feature_extractor,
    nn.Linear(...)
)
# model[0].parameters() and feature_extractor.parameters() are pointers to the same tensors.

kfac = KFACLinearOperator(model_func=model, ...)
kfac.load_state_dict(torch.load("kfac_state_dict.torch"))
# model[0].parameters() and feature_extractor.parameters() are now pointers to different tensors.
```

**2)** is an issue that we've already unexpectedly run into.

---

Proposed solution: Do not overwrite the `state_dict` of the model passed to `KFACLinearOperator` when loading the `state_dict` into a linear operator. Instead, verify that the state_dicts are the same, and throw an error if they're not (it should be up to the user to make sure the `state_dict` of the model matches first). Hence, `self._model_func` is still guaranteed to, numerically, be as expected, but we avoid keeping a new copy of `state_dict` in memory with different pointers.

### DIFF
--- a/curvlinops/utils.py
+++ b/curvlinops/utils.py
@@ -1,6 +1,6 @@
 """General utility functions."""
 
-from typing import List, Tuple, Union
+from typing import Any, List, Tuple, Union
 
 from numpy import cumsum
 from torch import Tensor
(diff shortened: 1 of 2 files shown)


runame (approved):
LGTM!

If you prefer avoiding the private functions you can just tell `ruff` to ignore the specific error, but I don't mind either way.

BrunoKM:
Happy to go ahead and merge then!

curvlinops/utils.py
```diff
@@ -115,3 +116,59 @@ def assert_divisible_by(num: int, divisor: int, name: str):
...
+    return True
+
+
+def _do_mappings_match(
+    statedict1: Mapping[str, StateDictType], statedict2: Mapping[str, StateDictType]
+) -> bool:
+    if len(statedict1) != len(statedict2):
+        return False
+    for key in statedict1.keys():
+        if key not in statedict2:
```
f-dangel, line 170: This does not catch the case where `statedict2`'s keys are the same as `statedict1`'s but it has *additional* keys.
Maybe better to return `False` if `set(statedict1.keys()) != set(statedict2.keys())`.

f-dangel (changes requested):
~One minor thing, otherwise lgtm!~ @BrunoKM could you also add an entry in the `changelog` file?

f-dangel:
Sorry I forgot one thing. Could we have a small test that reproduces the bug to ensure we do not accidentally re-introduce it in the future? Something like @BrunoKM described in the initial post of this issue.

curvlinops/utils.py
runame: I think this is not possible since the length has to be equal, see [here](https://github.com/f-dangel/curvlinops/pull/172/files#diff-6c9b95e40f318a96901d26a1f8b3f64b79687795687a4913c8326d4a545bd7f0R167-R168).

f-dangel: Thanks, makes sense!

Issue #126, opened 2024-09-17
[REF] Create shared parent classes for `KFACLinearOperator` and `KFACInverseLinearOperator`


runame:
I think we should at some point refactor `KFACLinearOperator` and `KFACInverseLinearOperator` to inherit from `KroneckerProductLinearOperator` and `EigendecomposedKroneckerProductLinearOperator` (or similar) classes since `torch_matmat` and other methods can be shared. Also, currently `KFACInverseLinearOperator` doesn't support trace, det, etc. properties which can also be shared.

Issue #116, opened 2024-05-25
Implement EKFAC


runame:
Implement Eigenvalue-corrected Kronecker Factorization [[George et al. (2018)](https://arxiv.org/abs/1806.03884)].

PR #173, opened 2025-02-20
[ADD] `EKFACLinearOperator`

runame:
Addresses #116 by adding an `EKFACLinearOperator` class (supersedes #127). Also refactors some of the KFAC tests such that their code can be shared for EKFAC tests.

This is a big diff and the code is not very polished yet, but I think it's good enough to receive feedback and start iterating.

Note that I restrict `EKFACLinearOperator` to only work with models with 2d output. This shifts the responsibility to the user to reshape according to their use case. The convention we use elsewhere (e.g. in `KFACLinearOperator`) cannot be applied here, because the per-example gradients can only be computed for _independent_ loss terms, i.e. the loss terms over data point (at least with the chosen implementation strategy).

**Adopted from #127**: I think we should at some point refactor `KFACLinearOperator` and `KFACInverseLinearOperator `to inherit from `KroneckerProductLinearOperator` and `EigendecomposedKroneckerProductLinearOperator` (or similar) classes since `_matmat` and other methods can be shared. Also, currently `KFACInverseLinearOperator` doesn't support trace, det, etc. properties which can also be shared. I created https://github.com/f-dangel/curvlinops/issues/126 for this.

Also, currently `KFACInverseLinearOperator` is used for `KFACLinearOperator` AND `EKFACLinearOperator`, which might be confusing, but makes sense in terms of the code. The refactor idea above could also help with this.

**Review order suggestion:**
- [x] 1. `curvlinops/kfac.py`
- [x] 2. `test/test_kfac.py`
- [x] 3. `curvlinops/inverse.py`
- [x] 4. `test/test_inverse.py`
- [x] 5. `curvlinops/ekfac.py`
- [x] 6. `test/test_ekfac.py` jointly with `test/utils.py`
- [x] 7. `curvlinops/__init__.py`


curvlinops/ekfac.py
```diff
@@ -0,0 +1,642 @@
...
+        return output, y
+
+    def _maybe_compute_ekfac(self):
+        """Compute the EKFAC approximation when necessary."""
+        if not self._corrected_eigenvalues:
+            if not (self._input_covariances or self._gradient_covariances):
+                self.compute_kronecker_factors()
+            self.compute_eigenvalue_correction()
+
+    def _matmat(self, M: List[Tensor]) -> List[Tensor]:
```
runame, line 200: There is a lot of code duplication here.

curvlinops/kfac.py
```diff
@@ -312,6 +312,90 @@ def _reset_matrix_properties(self):
...
+            M_joint = einsum(
+                ggT_eigvecs, M_joint, aaT_eigvecs, "i j, i k v, k l -> j l v"
+            )
+            # Multiply (broadcasted) by eigenvalues.
+            M_joint.mul_(eigenvalues.unsqueeze(-1))
+            # Transform back to standard basis.
+            M_joint = einsum(
+                ggT_eigvecs, M_joint, aaT_eigvecs, "i j, j k v, l k -> i l v"
+            )
+        return M_joint
```
f-dangel, line 353: Could rename `M_joint` into `M` since the `_joint` does not add much meaning here.
We should document that `M` is a 3d tensor: the first axis matches the first Kronecker matrix, the second axis the second Kronecker matrix, and the last is the free dimension for carrying out multiple multiplies in parallel.

curvlinops/kfac.py
```diff
@@ -312,6 +312,90 @@ def _reset_matrix_properties(self):
...
+            ggT: Gradient covariance Kronecker factor or its eigenvectors.
+            eigenvalues: Eigenvalues of the (E)KFAC approximation when multiplying with
+                the eigendecomposition of the KFAC approximation. ``None`` for the
+                non-decomposed KFAC approximation. Defaults to ``None``.
+        """
+        for p_name, pos in param_pos.items():
+            # for weights we need to multiply from the right with aaT
+            # for weights and biases we need to multiply from the left with ggT
+            if p_name == "weight":
+                M_w = rearrange(M[pos], "m ... v -> m (...) v")
```
f-dangel, line 381: `M[pos]` hints that the type annotation of `M` should also be `List[Tensor]` (but it is `Tensor` at the moment). Can you also update the docstring accordingly?

curvlinops/kfac.py
```diff
@@ -312,6 +312,90 @@ def _reset_matrix_properties(self):
...
+                # If `eigenvalues` is not `None`, we transform to eigenbasis here
+                KM[pos] = einsum(M_w, aaT, "c_out j v, j k -> c_out k v")
```
f-dangel, line 383: Can we keep the `einsum` indices for the first dimension of `M_w` consistent across these two `einsum`s? I would prefer `c_out`.

curvlinops/kfac.py
```diff
@@ -312,6 +312,90 @@ def _reset_matrix_properties(self):
...
+                KM[pos] = M[pos]
+
+            # If `eigenvalues` is not `None`, we convert to eigenbasis here
+            KM[pos] = einsum(
+                ggT.T if eigenvalues else ggT, KM[pos], "j k, k ... v -> j ... v"
+            )
+
+            if eigenvalues is not None:
+                # Multiply (broadcasted) by eigenvalues, convert back to original basis
+                KM[pos].mul_(eigenvalues[pos].unsqueeze(-1))
```
f-dangel, line 394: This suggests that `eigenvalues`'s type annotation should be `List[Tensor]`.

curvlinops/kfac.py
```diff
@@ -312,6 +312,90 @@ def _reset_matrix_properties(self):
...
+                if p_name == "weight":
+                    KM[pos] = einsum(KM[pos], aaT, "c_out j v, k j -> c_out k v")
+                KM[pos] = einsum(ggT, KM[pos], "j k, k ... v -> j ... v")
```
f-dangel, line 397: Why don't we have to `unflatten` the `...` dimension? It seems that `M[pos]` and `KM[pos]` have different shapes, but they should be identical.

f-dangel: Update: I see this is done further below outside of the routine. Will leave a comment below.

curvlinops/kfac.py
```diff
@@ -346,37 +434,22 @@ def _matmat(self, M: List[Tensor]) -> List[Tensor]:
             ):
                 w_pos, b_pos = param_pos["weight"], param_pos["bias"]
                 # v denotes the free dimension for treating multiple vectors in parallel
-                M_w = rearrange(M[w_pos], "c_out ... v -> c_out (...) v")
+                M_w = rearrange(M[w_pos], "m ... v -> m (...) v")
```
f-dangel, line 437: I think this line allows us to remove the same `if` condition in `_left_and_right_multiply` [above](https://github.com/f-dangel/curvlinops/pull/173/files#diff-81987851ce797c96eec5b13c693f28ffd852d7a3af0ab1e712d5da58a02a9ed9R380-R381).

f-dangel: Note: We can also replace some `...` in `_left_and_right_multiply` with a single letter, which should be safer.

curvlinops/kfac.py
```diff
@@ -413,12 +486,38 @@ def _compute_kfac(self):
...
             self._compute_loss_and_backward(output, y)
 
         # clean up
         for handle in hook_handles:
             handle.remove()
 
+    def _rearrange_for_larger_than_2d_output(
+        self,
+        output: Tensor,
+        y: Tensor,
```
f-dangel, line 499: Nit: Might fit into a single line.

curvlinops/kfac.py
```diff
@@ -346,37 +434,22 @@ def _matmat(self, M: List[Tensor]) -> List[Tensor]:
...
+                self._separate_left_and_right_multiply(KM, M, param_pos, aaT, ggT)
 
             # restore original shapes
             if "weight" in param_pos:
                 KM[param_pos["weight"]] = KM[param_pos["weight"]].view(weight_shape)
 
         return KM
 
-    def _compute_kfac(self):
+    def compute_kronecker_factors(self):
```
f-dangel, line 452: Is this method public on purpose now?

f-dangel (commented):
Did a pass over `kfac.py`.

curvlinops/kfac.py
runame: Yes, I believe it makes sense to expose this to the user. For example, imagine the setting where you want to fit EKFAC in two steps: 1) first, you set `check_deterministic=False`, 2) you fit the Kronecker factors with `compute_kronecker_factors` and save to disk, 2) you load from disk in a different job and compute the eigenvalue corrections.

curvlinops/kfac.py
runame: My understanding is that `KM[pos]` only has two dimensions when `p_name="bias"` and three dimensions when `p_name="weight'`; hence why we need `...`.

curvlinops/kfac.py
runame: Unless I'm missing something the line appears twice because they are in separate code paths: the line above is called before `_left_and_right_multiply`. However, the line you [reference](https://github.com/f-dangel/curvlinops/pull/173/files#diff-81987851ce797c96eec5b13c693f28ffd852d7a3af0ab1e712d5da58a02a9ed9R380-R381) is in `_separate_left_and_right_multiply`.

> Note: We can also replace some ... in _left_and_right_multiply with a single letter, which should be safer.

`_left_and_right_multiply` doesn't have any `...` and I think the `...` in `_separate_left_and_right_multiply` are necessary because of what I said in my comment [here](https://github.com/f-dangel/curvlinops/pull/173#discussion_r1965325561).

test/test_kfac.py
```diff
@@ -340,23 +423,37 @@ def test_kfac_mc_one_datum(
...
 def test_kfac_ef_one_datum(
     kfac_exact_one_datum_case: Tuple[
         Module,
         Union[BCEWithLogitsLoss, CrossEntropyLoss],
         List[Parameter],
         Iterable[Tuple[Tensor, Tensor]],
     ],
+    separate_weight_and_bias: bool,
+    exclude: str,
 ):
```
f-dangel, line 441: I wonder if the functionality inside this test can be refactored into a single function that is shared by all tests. This would reduce a lot of code duplication, e.g. `if exclude is not None: ...` condition.

f-dangel (commented):
Took a look at `test/test_kfac.py` and think there is some potential for refactoring.

test/test_kfac.py
f-dangel: This is a bit of a tangent to the current PR, but since we are introducing a lot of new complexity, it might be worth reduce it elsewhere.

curvlinops/inverse.py
```diff
@@ -429,195 +485,101 @@ def _compute_inverse_factors(
...
+            biases).
         """
+        if isinstance(self._A, EKFACLinearOperator):
+            aaT_eigenvectors = self._A._input_covariances_eigenvectors.get(name)
+            ggT_eigenvectors = self._A._gradient_covariances_eigenvectors.get(name)
+            eigenvalues = self._A._corrected_eigenvalues[name]
+            if isinstance(eigenvalues, dict):
+                inv_damped_eigenvalues: Dict[str, Tensor] = {}
+                for key, val in eigenvalues.items():
+                    inv_damped_eigenvalues[key] = val.add(self._damping).pow_(-1)
```
f-dangel, line 542: Nit: Could write as single dict comprehension.

curvlinops/inverse.py
```diff
@@ -649,20 +619,22 @@ def _matmat(self, X: List[Tensor]) -> List[Tensor]:
...
                 X_joint = cat([X_w, X[b_pos].unsqueeze(1)], dim=1)
-                X_joint = self._left_and_right_multiply(X_joint, aaT_inv, ggT_inv)
+                X_joint = self._A._left_and_right_multiply(
+                    X_joint, aaT_inv, ggT_inv, inv_damped_eigenvalues
+                )
                 w_cols = X_w.shape[1]
-                X[w_pos], X[b_pos] = X_joint.split([w_cols, 1], dim=1)
-                X[b_pos] = X[b_pos].squeeze(1)
+                KX[w_pos], KX[b_pos] = X_joint.split([w_cols, 1], dim=1)
+                KX[b_pos] = KX[b_pos].squeeze(1)
```
f-dangel, line 627: Nit: Could just do `KX[b_pos].squeeze_()`.

f-dangel (commented):
`inverse.py` lgtm up to some nits!

test/test_inverse.py
```diff
@@ -663,12 +667,179 @@ def test_KFAC_inverse_from_state_dict(
...
+
+    # use exact damping with KFACInverseLinearOperator
+    inv_EKFAC = KFACInverseLinearOperator(
+        EKFAC, damping=delta, cache=cache, use_exact_damping=True
+    )
+
+    num_vectors = 2
+    X = random.rand(EKFAC.shape[1], num_vectors)
+    # test for equivalence
+    report_nonclose(inv_EKFAC.to_scipy() @ X, inv_EKFAC_naive @ X)
```
f-dangel, line 757: Could be better to use PyTorch quantities throughout the tests and not mix SciPy with PyTorch.
Now that almost all operators are purely PyTorch, I would like to eliminate as much SciPy contamination as possible. Do you agree?

test/test_inverse.py
```diff
@@ -663,12 +667,179 @@ def test_KFAC_inverse_from_state_dict(
...
+    batch_size, D_in, D_out = 4, 3, 2
+    X = torch.rand(batch_size, D_in)
+    y = torch.rand(batch_size, D_out)
+    model = torch.nn.Linear(D_in, D_out)
+
+    params = list(model.parameters())
+    # create and compute EKFAC
+    ekfac = EKFACLinearOperator(
+        model,
+        MSELoss(reduction="sum"),
```
f-dangel, line 777: Any specific reason for using a non-default argument for `reduction` here?

test/test_inverse.py
```diff
@@ -663,12 +667,179 @@ def test_KFAC_inverse_from_state_dict(
...
+        params,
+        [(X, y)],
+    )
+
+    # create inverse KFAC
+    inv_ekfac = KFACInverseLinearOperator(
+        ekfac, damping=1e-2, use_exact_damping=True, retry_double_precision=False
+    )
+    _ = inv_ekfac.to_scipy() @ eye(ekfac.shape[1])  # to trigger inverse computation
```
f-dangel, line 786: Same thought about SciPy as above.

test/test_inverse.py
```diff
@@ -663,12 +667,179 @@ def test_KFAC_inverse_from_state_dict(
...
+            torch.load("inv_ekfac_state_dict.pt", weights_only=False)
+        )
+
+    # create new inverse KFAC and load state dict
+    inv_ekfac_new = KFACInverseLinearOperator(ekfac, use_exact_damping=True)
+    inv_ekfac_new.load_state_dict(
+        torch.load("inv_ekfac_state_dict.pt", weights_only=False)
+    )
+    # clean up
+    os.remove("inv_ekfac_state_dict.pt")
```
f-dangel, line 808: Refactor path for `.pt` file into a variable.

test/test_inverse.py
```diff
@@ -663,12 +667,179 @@ def test_KFAC_inverse_from_state_dict(
...
+    batch_size, D_in, D_out = 4, 3, 2
+    X = torch.rand(batch_size, D_in)
+    y = torch.rand(batch_size, D_out)
+    model = torch.nn.Linear(D_in, D_out)
+
+    params = list(model.parameters())
+    # create and compute EKFAC
+    ekfac = EKFACLinearOperator(
+        model,
+        MSELoss(reduction="sum"),
```
f-dangel, line 828: Same as above.

f-dangel (commented):
Minor refactoring suggestions for `test/test_inverse.py`.

curvlinops/ekfac.py
```diff
@@ -0,0 +1,638 @@
...
+
+            # Compute eigendecomposition of the Kronecker factors and cache eigenvectors
+            if aaT is not None:
+                _, aaT_eigvecs = eigh(aaT)
+                self._input_covariances_eigenvectors[mod_name] = aaT_eigvecs
+                del aaT
+            if ggT is not None:
+                _, ggT_eigvecs = eigh(ggT)
+                self._gradient_covariances_eigenvectors[mod_name] = ggT_eigvecs
+                del ggT
```
f-dangel, line 276: Refactor into a loop:
```python
for source, destination in zip(
    (self._input_covariances, self._gradient_covariances), 
    (self._input_covariances_eigenvectors, self._gradient_covariances_eigenvectors)
):
    factor = source.pop(mod_name, None)
    if factor is not None:
        destination[mod_name] = eigh(factor)[1]
        # I think we don't need the `del` statement any more because `factor` will be overwritten
        # in the next iteration of the for loop and this will decrease the ref count of the previous 
        # factor to zero, which should trigger garbage collection
```

curvlinops/ekfac.py
```diff
@@ -0,0 +1,638 @@
...
+        correction = {
+            "sum": 1.0 / self._mc_samples,
+            "mean": num_loss_terms**2
+            / (self._N_data * self._mc_samples * self._num_per_example_loss_terms),
+        }[self._loss_func.reduction]
+
+        # Compute the corrected eigenvalues for the EKFAC approximation
+        param_pos = self._mapping[module_name]
+        aaT_eigenvectors = self._input_covariances_eigenvectors.get(module_name)
+        ggT_eigenvectors = self._gradient_covariances_eigenvectors.get(module_name)
```
f-dangel, line 416: Better to use `[module_name]` here because this may still pass if the key does not exist?

curvlinops/ekfac.py
```diff
@@ -0,0 +1,638 @@
...
+    def load_state_dict(self, state_dict: Dict[str, Any]):
+        """Load the state of the EKFAC linear operator.
+
+        Args:
+            state_dict: State dictionary.
+        """
+        super().load_state_dict(state_dict)
+
+        # Set EKFAC-specific quantities
+        # TODO: should we check if the keys match the mapping keys?
```
f-dangel, line 630: I think this would be good because this will allow detecting inconsistencies when loading a state dict that would otherwise lead to errors only much later (e.g. in a matvec).

f-dangel (commented):
I honestly do not have many suggestions to improve `ekfac.py`. 
The code duplication seems necessary for me, because we do not have a good abstraction for multiplying Kronecker-factored (and potentially eigendecomposed) matrices.

test/test_kfac.py
runame: Yes, I agree that there is tons of code duplication in the tests that could be reduced. I can maybe do some simple refactoring by just moving highly duplicated code into separate functions, as suggested, but leave more comprehensive refactoring for later.

test/test_inverse.py
runame: I agree, the only thing that is a bit awkward is that `report_nonclose` only works with numpy arrays, so we always have to convert to numpy first. I just do this for now, but we could maybe consider extending `report_nonclose` to also work with torch tensors.

test/test_inverse.py
runame: To make sure the reduction is correctly saved and loaded. I added a comment.

test/test_kfac.py
runame: See [this commit](https://github.com/f-dangel/curvlinops/pull/173/commits/183d06feaed907e3c8a1a464608c3f99f8d2580d).

curvlinops/ekfac.py
runame: We can only do this for `ggT_eigenvectors` because `aaT_eigenvectors` does not exist when the weight matrix is excluded. I made the change for `ggT_eigenvectors` and added a clarifying comment.

test/utils.py
```diff
@@ -92,36 +93,38 @@ def regression_targets(size: Tuple[int]) -> Tensor:
...
-def ggn_block_diagonal(
+def block_diagonal(
+    linear_operator: PyTorchLinearOperator,
     model: Module,
     loss_func: Module,
     params: List[Parameter],
     data: Iterable[Tuple[Union[Tensor, MutableMapping], Tensor]],
     batch_size_fn: Optional[Callable[[MutableMapping], int]] = None,
     separate_weight_and_bias: bool = True,
 ) -> ndarray:
```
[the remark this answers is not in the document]
f-dangel: See #174.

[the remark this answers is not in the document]
f-dangel: See #174

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        loss_func,
+        params,
+        data,
+        batch_size_fn=batch_size_fn,
+        fisher_type=FisherType.TYPE2,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
+
+    report_nonclose(ggn, ekfac_mat, atol=1e-6)
```
f-dangel, line 101: Let's use PyTorch and `assert allclose_report(...)`

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+            or ``None``.
+        separate_weight_and_bias: Whether to treat weight and bias as separate blocks in
+            the EKFAC matrix.
+    """
+    model, loss_func, params, data, batch_size_fn = kfac_weight_sharing_exact_case
+    # The model outputs have to be flattened assuming only the first dimension is the
+    # batch dimension since EKFAC only supports 2d outputs.
+    model.setting = "expand-flatten" if "expand" in setting else setting
+    if isinstance(model, Conv2dModel):
+        # parameters are only initialized after the setting property is set
```
f-dangel, line 146: Not sure I understand how the comment connects to the next line of code here.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        params,
+        data,
+        batch_size_fn=batch_size_fn,
+        fisher_type=FisherType.TYPE2,
+        kfac_approx=setting,  # choose EKFAC approximation consistent with setting
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
+
+    report_nonclose(ggn, ekfac_mat, rtol=1e-4)
```
f-dangel, line 175: Same as above.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        fisher_type=FisherType.MC,
+        mc_samples=2_000,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
+
+    atol = {"sum": 5e-1, "mean": 5e-3}[loss_func.reduction]
+    rtol = {"sum": 2e-2, "mean": 2e-2}[loss_func.reduction]
+
+    report_nonclose(ggn, ekfac_mat, rtol=rtol, atol=atol)
```
f-dangel, line 235: Same as above.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+            ``KFACType.REDUCE``.
+        shuffle: Whether to shuffle the parameters before computing the EKFAC matrix.
+    """
+    model, loss_func, params, data, batch_size_fn = kfac_weight_sharing_exact_case
+    # The model outputs have to be flattened assuming only the first dimension is the
+    # batch dimension since EKFAC only supports 2d outputs.
+    model.setting = "expand-flatten" if "expand" in setting else setting
+    if isinstance(model, Conv2dModel):
+        # parameters are only initialized after the setting property is set
+        params = [p for p in model.parameters() if p.requires_grad]
```
f-dangel, line 275: Same as above.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        check_deterministic=False,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
+
+    # Scale absolute tolerance by the number of outputs when using sum reduction.
+    num_outputs = sum(y.numel() for _, y in data)
+    atol = {"sum": 5e-3 * num_outputs, "mean": 5e-3}[loss_func.reduction]
+    rtol = {"sum": 2e-2, "mean": 2e-2}[loss_func.reduction]
+
+    report_nonclose(ggn, ekfac_mat, rtol=rtol, atol=atol)
```
f-dangel, line 310: Same as above.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        loss_func,
+        params,
+        data,
+        batch_size_fn=batch_size_fn,
+        fisher_type=FisherType.TYPE2,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
+
+    report_nonclose(ggn, ekfac_mat)
```
f-dangel, line 354: Same as above.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        fisher_type=FisherType.MC,
+        mc_samples=11_000,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
+
+    atol = {"sum": 1e-3, "mean": 1e-3}[loss_func.reduction]
+    rtol = {"sum": 3e-2, "mean": 3e-2}[loss_func.reduction]
+
+    report_nonclose(ggn, ekfac_mat, rtol=rtol, atol=atol)
```
f-dangel, line 402: Same as above.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        loss_func,
+        params,
+        data,
+        batch_size_fn=batch_size_fn,
+        fisher_type=FisherType.EMPIRICAL,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
+
+    report_nonclose(ef, ekfac_mat, atol=1e-7)
```
f-dangel, line 447: Same as above.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+
+    report_nonclose(ggn, ekfac_mat, rtol=rtol, atol=atol)
+
+    # 2) Compare GGN (inplace=True) and GGN (inplace=False)
+    for mod in model.modules():
+        if hasattr(mod, "inplace"):
+            mod.inplace = False
+    ggn_no_inplace = block_diagonal(GGNLinearOperator, model, loss_func, params, data)
+
+    report_nonclose(ggn, ggn_no_inplace)
```
f-dangel, line 484: This test should be refactored to share the same code as the equivalent test for KFAC.

f-dangel (commented):
Made it halfway through `test/test_ekfac.py`.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+    ekfac_mean = EKFACLinearOperator(
+        model,
+        loss_func,
+        params,
+        data,
+        fisher_type=fisher_type,
+    )
+    ekfac_mean_mat = ekfac_mean.to_scipy() @ eye(ekfac_mean.shape[1])
+
+    report_nonclose(ekfac_simulated_mean_mat, ekfac_mean_mat, atol=1e-4)
```
f-dangel, line 629: Use PyTorch.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+
+    # Check for equivalence of trace property and naive trace computation
+    trace = ekfac.trace
+    trace_naive = (ekfac.to_scipy() @ eye(ekfac.shape[1])).trace()
+    report_nonclose(trace.cpu().numpy(), trace_naive)
+
+    # Check that the trace property is properly cached and reset
+    assert ekfac._trace == trace
+    ekfac.compute_kronecker_factors()
+    assert ekfac._trace is None
```
f-dangel, line 669: This test should share code with its KFAC equivalent.

f-dangel: Same for the Frobenius norm test and det, logdet.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+                eigenvals.add_(delta)
+        else:
+            eigenvalues.add_(delta)
+
+    # Check for equivalence of the det property and naive determinant computation
+    determinant = ekfac.det
+    # verify that the determinant is not trivial as this would make the test useless
+    assert determinant not in {0.0, 1.0}
+    det_naive = det(ekfac.to_scipy() @ eye(ekfac.shape[1]))
+    report_nonclose(determinant.cpu().numpy(), det_naive, rtol=1e-4)
```
f-dangel, line 757: Use PyTorch.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        else:
+            eigenvalues.add_(delta)
+
+    # Check for equivalence of the logdet property and naive log determinant computation
+    log_det = ekfac.logdet
+    # verify that the log determinant is finite and not nan
+    assert not isinf(log_det) and not isnan(log_det)
+    sign, logabsdet = slogdet(ekfac.to_scipy() @ eye(ekfac.shape[1]))
+    log_det_naive = sign * logabsdet
+    report_nonclose(log_det.cpu().numpy(), log_det_naive, rtol=1e-4)
```
f-dangel, line 814: Use PyTorch.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+
+    # Check that the logdet property is properly cached and reset
+    assert ekfac._logdet == log_det
+    ekfac.compute_kronecker_factors()
+    assert ekfac._logdet is None
+
+
+def test_ekfac_does_affect_grad():
```
f-dangel, line 822: Missing `not` in function name?

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+    # save state dict
+    state_dict = ekfac.state_dict()
+
+    # create new EKFAC from state dict
+    kfac_new = EKFACLinearOperator.from_state_dict(state_dict, model, params, [(X, y)])
+
+    # check that the two EKFACs are equal
+    compare_state_dicts(ekfac.state_dict(), kfac_new.state_dict())
+    test_vec = rand(ekfac.shape[1])
+    report_nonclose(ekfac @ test_vec, kfac_new @ test_vec)
```
f-dangel, line 952: Could these tests also share code with their KFAC equivalents?

f-dangel: Could be worth to refactor them into a more general function that tests whether a stateful linear operator can be successfully saved and loaded from its state dict.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        model,
+        loss_func,
+        params,
+        data,
+        batch_size_fn=batch_size_fn,
+        separate_weight_and_bias=separate_weight_and_bias,
+        fisher_type=fisher_type,
+        mc_samples=1_000 if fisher_type == FisherType.MC else 1,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
```
f-dangel, line 1010: Use PyTorch.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+    separate_weight_and_bias: bool,
+    fisher_type: FisherType,
+):
+    """Test that EKFAC is closer in Frobenius norm to the exact quantity than KFAC."""
+    model, loss_func, params, data, batch_size_fn = inv_case
+    params = maybe_exclude_or_shuffle_parameters(params, model, exclude, shuffle)
+
+    # Compute exact block-wise ground truth quantity.
+    linop = (
+        EFLinearOperator if fisher_type == FisherType.EMPIRICAL else GGNLinearOperator
```
f-dangel, line 976: What if the Fisher type is type-I MC?

In this case, EKFAC should be a better approximation to the MC Fisher than KFAC, and so the linop (the 'exact' quantity) would have to be `FisherMCLinearOperator` with the same seed as (E)KFAC.

f-dangel: Then we could always set `mc_samples=1` below.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+    exact_ekfac_dist = norm(exact - ekfac_mat) / exact_norm
+    if exclude == "weight":
+        # For no_weights the numerical error dominates.
+        assert np_allclose(exact_kfac_dist, exact_ekfac_dist, atol=1e-5)
+    else:
+        assert exact_kfac_dist > exact_ekfac_dist
+
+
+# TODO: FisherType.MC is too expensive, but could be added as optional test?
+@mark.parametrize("fisher_type", [FisherType.TYPE2, FisherType.EMPIRICAL])
```
f-dangel, line 1024: Same comment as above. Should we just use the MC-Fisher with the identical random seed as exact quantity? Then, the test should pass without having to crank up the number of MC samples.

test/test_ekfac.py
```diff
@@ -0,0 +1,1094 @@
...
+        model,
+        loss_func,
+        params,
+        data,
+        batch_size_fn=batch_size_fn,
+        separate_weight_and_bias=separate_weight_and_bias,
+        fisher_type=fisher_type,
+        kfac_approx=kfac_approx,
+    )
+    ekfac_mat = ekfac.to_scipy() @ eye(ekfac.shape[1])
```
f-dangel, line 1084: Use PyTorch.

f-dangel (commented):
Full pass through EKFAC tests.

test/test_inverse.py
runame: Oh cool, I didn't realise this exists. I used it in this and a few other instances in [this commit](https://github.com/f-dangel/curvlinops/pull/173/commits/4a5230938e77e8eb258b0bbc5321d74fe13d1991).

test/utils.py
```diff
@@ -250,7 +255,11 @@ def forward(self, x: Tensor) -> Tensor:
         # Example: Transformer for translation: (batch, sequence_length, c)
         # (although second and third dimension would have to be transposed for
         # classification)
-        if x.ndim > 2 and self.loss == "CE":
+        if self.setting == "expand-flatten":
+            # Example: Per-sequence loss for transformer with outputs:
+            # (batch, sequence_length, c)
```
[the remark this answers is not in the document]
runame: Replace this with a better example.

test/test_ekfac.py
runame: I clarified the comment.

test/test_ekfac.py
runame: Same for KFAC actually, nice catch.

test/test_ekfac.py
runame: That's a great idea, thanks!

test/test_ekfac.py
```diff
@@ -0,0 +1,949 @@
...
+    ekfac = EKFACLinearOperator(
+        model,
+        loss_func,
+        params,
+        data,
+        batch_size_fn=batch_size_fn,
+        fisher_type=FisherType.TYPE2,
+        separate_weight_and_bias=separate_weight_and_bias,
+    )
+    ekfac_mat = ekfac @ eye(ekfac.shape[1])
```
f-dangel, line 96: This should crash if `ekfac` lives on GPU.

f-dangel: There are a couple more of these cases in this file.

test/utils.py
```diff
@@ -633,3 +705,208 @@ def relative_l_inf_error(a_true: ndarray, a: ndarray) -> float:
...
+        GGNLinearOperator, model, loss_func, params, data, return_numpy=False
+    )
+
+    linop = linop_cls(model, loss_func, params, data, mc_samples=2_000)
+    linop_mat = linop @ eye(linop.shape[1])
+
+    atol = {"sum": 5e-1, "mean": 2e-3}[loss_func.reduction]
+    rtol = {"sum": 2e-2, "mean": 2e-2}[loss_func.reduction]
+
+    assert allclose_report(ggn, linop_mat, rtol=rtol, atol=atol)
```
f-dangel, line 740: Would it be better to test against type-II instead? This uses less compute than 2000 MC samples, and we can also use smaller tolerances.

test/utils.py
```diff
@@ -633,3 +705,208 @@ def relative_l_inf_error(a_true: ndarray, a: ndarray) -> float:
...
+    model: Module,
+    loss_func: Module,
+    params: List[Parameter],
+    data: Iterable[Tuple[Union[Tensor, MutableMapping], Tensor]],
+    batch_size_fn: Optional[Callable[[MutableMapping], int]],
+    separate_weight_and_bias: bool,
+    check_deterministic: bool,
+    rtol: float = 1e-5,
+    atol: float = 1e-8,
+):
```
f-dangel, line 765: This currently does not test different KFAC approximations and Fisher types. Is that okay?

test/utils.py
```diff
@@ -633,3 +705,208 @@ def relative_l_inf_error(a_true: ndarray, a: ndarray) -> float:
...
+        check_deterministic=check_deterministic,
+    )
+    # Mapping from the property name to the corresponding torch function
+    torch_fn_map = {
+        "trace": trace,
+        "frobenius_norm": linalg.matrix_norm,
+        "det": linalg.det,
+        "logdet": logdet,
+    }
+    assert property_name in torch_fn_map
```
f-dangel, line 800: Could just do `torch_fn = {...}[property_name]` and remove the `assert`.

test/utils.py
```diff
@@ -633,3 +705,208 @@ def relative_l_inf_error(a_true: ndarray, a: ndarray) -> float:
...
+            for eigenvalues in linop._corrected_eigenvalues.values():
+                if isinstance(eigenvalues, dict):
+                    for eigenvals in eigenvalues.values():
+                        eigenvals.add_(DELTA)
+                else:
+                    eigenvalues.add_(DELTA)
+
+    # Check for equivalence of property and naive computation
+    quantity = getattr(linop, property_name)
+    linop_mat = linop @ eye(linop.shape[1])
```
f-dangel, line 827: Should crash when `linop` lives on GPU.

test/utils.py
```diff
@@ -633,3 +705,208 @@ def relative_l_inf_error(a_true: ndarray, a: ndarray) -> float:
...
+    assert allclose_report(quantity, quantity_naive, rtol=rtol, atol=atol)
+
+    # Check that the property is properly cached and reset
+    assert getattr(linop, "_" + property_name) == quantity
+    linop.compute_kronecker_factors()
+    assert getattr(linop, "_" + property_name) is None
+
+
+def _test_save_and_load_state_dict(
+    lino_cls: Type[Union[KFACLinearOperator, EKFACLinearOperator]],
```
f-dangel, line 838: Rename into `linop_cls`

test/test_ekfac.py
f-dangel: Should be refactored like [here](https://github.com/f-dangel/curvlinops/pull/173/files#diff-7a31260cb833244158c8ec94421e00372de13b3d7fca3ccc37785e21647ba73cR897-R901), no?

f-dangel (commented):
One more round of suggestions.

test/test_ekfac.py
runame: I actually also tried to use the same trick here, but it doesn't work. Maybe the eigendecompositions are not stable enough for these cases?

test/test_ekfac.py
runame: I extended `eye_like` to also work with linops and use it everywhere in this file. In the future when we remove `to_scipy()` calls elsewhere we can also make use of it.

test/utils.py
runame: Yeah, no idea why MC samples are used here -- it's pretty much instant anyway, but we can get rid of the custom tolerances.

test/utils.py
runame: I think so since how these properties are computed only depends on the Kronecker factors and not on how they were computed.

docs/rtd/linops.rst
```diff
@@ -21,7 +21,10 @@ Fisher (approximate)
    :members: __init__
 
 .. autoclass:: curvlinops.KFACLinearOperator
-   :members: __init__, trace, det, logdet, frobenius_norm
+   :members: __init__, trace, det, logdet, frobenius_norm, state_dict, load_state_dict, from_state_dict
+
+.. autoclass:: curvlinops.EKFACLinearOperator
+   :members: __init__, trace, det, logdet, frobenius_norm, state_dict, load_state_dict, from_state_dict
```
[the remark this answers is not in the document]
runame: Yes, I think that makes sense. Is this the right syntax?
```rst
.. autoclass:: curvlinops.FisherType
   :members: TYPE2, MC, EMPIRICAL, FORWARD_ONLY

.. autoclass:: curvlinops.KFACType
   :members: EXPAND, REDUCE
```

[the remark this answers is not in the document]
f-dangel: You can check the output of the GH action that builds the RTD website (https://curvlinops--173.org.readthedocs.build/en/173/linops.html).

test/test_ekfac.py
```diff
@@ -0,0 +1,949 @@
...
+    # save state dict
+    state_dict = ekfac.state_dict()
+
+    # create new EKFAC from state dict
+    kfac_new = EKFACLinearOperator.from_state_dict(state_dict, model, params, [(X, y)])
+
+    # check that the two EKFACs are equal
+    compare_state_dicts(ekfac.state_dict(), kfac_new.state_dict())
+    test_vec = rand(ekfac.shape[1])
+    assert allclose_report(ekfac @ test_vec, kfac_new @ test_vec)
```
f-dangel, line 800: Can this test be shared with the equivalent KFAC test?

test/test_ekfac.py
f-dangel: Did you pass the same `seed` to `FisherMCLinearOperator` than to `EKFACLinerOperator` and `KFACLinearOperator`?

runame: Yes, I did. I noticed that only a single case was failing and modified it to have 3 instead of 2 classes (it was a CE loss case). After this adjustment, it seems stable. See the refactor [here](https://github.com/f-dangel/curvlinops/pull/173/commits/b1a20748011a3b861fa121dcff237c4f3dce36b8).

docs/rtd/linops.rst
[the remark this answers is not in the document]
runame: I tried the above and
```rst
.. autoclass:: curvlinops.FisherType
   :members:
   :member-order: bysource

.. autoclass:: curvlinops.KFACType
   :members:
   :member-order: bysource
```
but neither worked, no idea why. I think it should in principle based on what I found online. Should we tackle this in a follow up PR since it is unrelated to the main topic of this one?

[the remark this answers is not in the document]
f-dangel: Please open an issue and we will not prioritize this in this PR.

f-dangel (commented):
LGTM now.

Remaining TODOs:
- [ ] Confirm tests are passing on GPU
- [ ] Add an entry to the changelog

docs/rtd/linops.rst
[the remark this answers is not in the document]
runame: Created an issue: #178.

PR #175, opened 2025-02-23
[ADD] Test two consecutive matvecs of a linear operator

### LINKED ISSUE #159: [TEST] Test that linear operator multiplication onto two identical objects yields the same result
This is currently verified implicitly for `CurvatureLinearOperator`s in their deterministic check, but does not include inverses like `KFACInverseLinearOperator`. As suggested by @runame, it will be good to add explicit tests for all linear operators.

f-dangel:
Resolves #159.

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -506,6 +506,44 @@ def compare_matmat(
     for o_x, m_x in zip(op_x, mat_x):
         assert allclose_report(o_x, m_x, **tol)
 
+def compare_consecutive_matmats(
+    op: PyTorchLinearOperator,
+    adjoint: bool,
+    is_vec: bool,
+    num_vecs: int = 2,
+    rtol: float = 1e-5,
+    atol: float = 1e-8,
+):
+    """Compare applying the linear operator to two identical vectors in sequence.
+
+    Args:
+        op: The operator to test.
+        adjoint: Whether to test the adjoint operator.
+        is_vec: Whether to test matrix-vector or matrix-matrix multiplication.
+        num_vecs: Number of vectors to test (ignored if ``is_vec`` is ``True``).
+            Default: ``2``.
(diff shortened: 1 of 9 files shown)


test/utils.py
```diff
@@ -507,6 +507,49 @@ def compare_matmat(
...
+    Args:
+        op: The operator to test.
+        adjoint: Whether to test the adjoint operator.
+        is_vec: Whether to test matrix-vector or matrix-matrix multiplication.
+        num_vecs: Number of vectors to test (ignored if ``is_vec`` is ``True``).
+            Default: ``2``.
+        rtol: Relative tolerance for the comparison. Default: ``1e-5``.
+        atol: Absolute tolerance for the comparison. Default: ``1e-8``.
+    """
+    if adjoint:
```
runame, line 529: Is there a reason to offer this `adjoint` option? I think it might be more explicit to just require the user to pass `op.adjoint()` to this function.

For example, when I first saw a call to this function with `adjoint=True`, I wasn't sure if it will compare the matmat for the linop AND its adjoint or only for the latter (which turns out to be the case).

test/utils.py
```diff
@@ -507,6 +507,49 @@ def compare_matmat(
...
+    atol: float = 1e-8,
+):
+    """Compare applying the linear operator to two identical vectors in sequence.
+
+    Args:
+        op: The operator to test.
+        adjoint: Whether to test the adjoint operator.
+        is_vec: Whether to test matrix-vector or matrix-matrix multiplication.
+        num_vecs: Number of vectors to test (ignored if ``is_vec`` is ``True``).
+            Default: ``2``.
```
runame, line 525: Is there a case where we want `num_vecs=1` and `is_vec=False`? If yes, why? Shouldn't it be sufficient to test things with `num_vec=1` and `is_vec=True` and then `num_vec=N` where `N>1`?

PR #180, opened 2025-04-18
Fix default value in `LSMRInverseLinearOperator`

### LINKED ISSUE #179: Misleading defaults in `LSMRInverseLinearOperator`?
The defaults for the LSMR hyperparameters (`conlim`) for `LSMRInverseLinearOperator` are different than for the underlying `scipy` function: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.lsmr.html

https://github.com/f-dangel/curvlinops/blob/b07fb257a8932b0aff8efa0d0f7307a91ff99b11/curvlinops/inverse.py#L116

The default for scipy is `1e8`.

Isn't the current default value for `conlim` of `1e-8` “technically” not a valid value? This should represent the maximum allowed condition number, and the condition number should not be smaller than 1. I might have missed something about the underlying algorithm, I'm just purely going off of the lsmr documentation.

runame:
Resolves #179.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   always carry out deterministic checks on the linear operator's device
   (previously always on CPU)
   ([PR](https://github.com/f-dangel/curvlinops/pull/160))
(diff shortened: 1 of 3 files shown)


f-dangel:
LGTM. Can you update the `changelog`, too?

f-dangel:
One more thing: Should we remove the `conlim=0` specifictation in the LSMR inverse tests?

f-dangel (approved):

PR #188, opened 2025-09-08
[REF] Translate trace and diagonal estimators to PyTorch

f-dangel:
Also found a small bug in XDiag that made the implementation work only on tensors, rather than linear operators.

TODO:
- [x] Update changelog


runame (approved):
LGTM

PR #196, opened 2025-10-15
Add warning about overwriting model parameters when loading state dict in (E)KFAC

runame:
When loading a state dict for `(E)KFACLinearOperator`, we overwrite the parameters of the model underlying the linear operator to correspond to the values used for the Kronecker factor computation (for consistency).

#172 raised the issue that this might result in unintended side effects. @f-dangel and me decided offline to still overwrite the model's parameters, but to add more explicit documentation and a warning via this PR (notably the more major cause for the suggestion in #172 appears to be caused by something else).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added/New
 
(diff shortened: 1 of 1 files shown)


f-dangel (approved):
LGTM.

PR #200, opened 2025-10-21
[REF] Implement multiplication with the GGN via `torch.func`

f-dangel:
Before, we used `BackPACK`'s GGN-vector products, which are based on nested reverse mode.
This PR refactors the multiplication to use `torch.func` which enables using forward mode.
Re-running the benchmark for the GGN, I observe significant performance improvements, specifically in peak memory (which is expected because we can use forward mode for the JVP, rather than double-backward, which is more memory-efficient):

- ResNet50: 13.3 GiB -> 6.2 GiB, 1.19s -> 0.69s
- NanoGPT: 18.2 GiB -> 11.0 GiB, 0.79s -> 0.78s

*Note:* I haven't tested with `torch.compile` yet but am certain this will further decrease memory and/or run time. This is somewhat orthogonal though.

### DIFF
--- a/curvlinops/ggn.py
+++ b/curvlinops/ggn.py
@@ -84,13 +84,13 @@ def make_batch_ggn_matrix_product(
     f, c = make_functional_model_and_loss(model_func, loss_func, params)
 
     # Create the functional GGN-vector product
-    ggn_vp = make_ggn_vector_product(f, c) # params, X, y, *v -> *Gv
+    ggn_vp = make_ggn_vector_product(f, c)  # params, X, y, *v -> *Gv
 
     # Fix the parameters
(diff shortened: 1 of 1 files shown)


curvlinops/ggn.py
```diff
@@ -1,12 +1,117 @@
...
+
+        Returns:
+            Result of GGN multiplication in list format. Has the same shape as
+            ``v``, i.e. each tensor in the list has the shape of a parameter.
+        """
+        return ggn_vp(params, X, y, *v)
+
+    # Vectorize over vectors to multiply onto a matrix in list format
+    return vmap(
+        ggn_vector_product,
```
runame: Is it possible to remove the whole `ggn_vector_product` function and replace this line with `functools.partial(ggn_vp, params=params)`?

PR #201, opened 2025-10-22
[REF] Implement multiplication with the Hessian via `torch.func`

f-dangel:
Same as #200, but for the Hessian.

Benchmark improvements (mostly memory reduction):
- ResNet50: 18.1 GiB -> 12.7 GiB, 1.27s -> 1.13s
- NanoGPT: 27.7 GiB -> 17.5 GiB, 2.94s -> 2.91s


runame (approved):
LGTM.

I have some questions about `torch.func`, but they can wait and we can discuss offline.

PR #202, opened 2025-10-23
[ADD] Implement empirical Fisher multiplication via `torch.func`

f-dangel:
Improvements:
- ResNet50: 13.3GiB -> 6.2GiB, 1.19s -> 0.67s
- NanoGPT: 18.2GiB -> 11.0GiB, 2.19s -> 2.09s

PR #204, opened 2025-10-30
[ADD] Implement MC-Fisher multiplication via `torch.func`

f-dangel:
Only review after #202 has been merged into this branch.

Improvements:
- ResNet50: 13.3 GiB -> 6.2 GiB, 1.20s -> 0.71s
- NanoGPT: 19.0 GiB -> 14.1 GiB, 2.38s -> 2.40s

For reference, the improvements for the empirical Fisher from #202 were:
- ResNet50: 13.3GiB -> 6.2GiB, 1.19s -> 0.67s
- NanoGPT: 18.2GiB -> 11.0GiB, 2.19s -> 2.09s

Some curious observations:
- On ResNet50+ImageNet I can barely see a memory difference between GGN, empirical, and MC Fisher (that's what I would have expected since all use GGN-vector products on a cheap-to-store pseudo-loss). However, on NanoGPT+Shakespeare there is a big gap in memory between MC and empirical Fisher/GGN, which is unexpected. Since sampling is the only difference, I did a quick search and found that there were large memory consumptions reported in the context of sampling large binomial tensors (https://github.com/pytorch/pytorch/issues/20343), but these issues seem to have been resolved. To further debug this, we could simply replace the Bernoulli sampling with generating an all-zeros tensor, then re-run the benchmark and see if that closes the gap, which would confirm that Bernoulli sampling is the issue. @runame during the review, could you try to check for any other differences between MC-Fisher and empirical Fisher/GGN that I might have missed? 
- It looks like the memory gap already existed before the refactoring, which supports my hypothesis that it might be related to sampling. The gap was less pronounced though due to our reliance on nested backprop.
- In terms of runtime, I observe that MC-Fisher is slightly slower than empirical Fisher/GGN, which I would have expected because sampling adds an overhead. The overhead looks okay, so at the moment I believe the problematic aspects are only related to memory.

### DIFF
--- a/curvlinops/fisher.py
+++ b/curvlinops/fisher.py
@@ -191,7 +191,7 @@ def make_batch_fmc_matrix_product(
         return 0.5 / (mc_samples * reduction_factor) * (inner_products**2).sum()
 
     # Create the functional MC Fisher-vector product using GGN of pseudo-loss
-    fmc_vp = make_ggn_vector_product(f_flat, c_pseudo_flat, c_extra_args=2)
+    fmc_vp = make_ggn_vector_product(f_flat, c_pseudo_flat, num_c_extra_args=2)
 
     # Fix the parameters: X, y, mc_samples, generator, *v -> *Fv
     fmcvp = partial(fmc_vp, params)
--- a/curvlinops/ggn.py
+++ b/curvlinops/ggn.py
@@ -13,7 +13,7 @@ from curvlinops.utils import make_functional_model_and_loss
 
 
 def make_ggn_vector_product(
-    f: Callable[..., Tensor], c: Callable[..., Tensor], c_extra_args: int = 0
+    f: Callable[..., Tensor], c: Callable[..., Tensor], num_c_extra_args: int = 0
 ) -> Callable[..., Tuple[Tensor, ...]]:
     """Create a function that computes GGN-vector products for given f and c functions.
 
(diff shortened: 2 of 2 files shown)


f-dangel:
I identified and fixed the culprit: I was not `detach`ing the NN prediction when computing pseudo-gradients. This built up the compute graph for sampling multinomials. Passing the `detach`ed value of the NN output to the sampling method fixes the issue:

Updated improvements:
- ResNet50: 13.3 GiB ->  6.2 GiB, 1.20s -> 0.71s
- NanoGPT: 19.0 GiB ->  11 GiB, 2.38s -> 2.25s

Everything looks as I would have expected now 👍

curvlinops/ggn.py
```diff
@@ -13,44 +13,53 @@
...
     """Create a function that computes GGN-vector products for given f and c functions.
 
     Args:
         f: Function that takes parameters and input, returns prediction.
             Signature: (*params, X) -> prediction
-        c: Function that takes prediction and target, returns loss.
-            Signature: (prediction, y) -> loss
+        c: Function that takes prediction, target, and optional additional args.
+            Signature: (prediction, y, *args) -> loss
+        c_extra_args: Number of additional arguments that the loss function c expects
```
runame: Nit: Based on the name I would expect the actual extra arguments for `c`, could be clarified by changing to something like `num_c_extra_args`.

curvlinops/fisher.py
```diff
@@ -1,16 +1,212 @@
...
+            mc_samples: Number of MC samples to use.
+            generator: Random generator for sampling.
+
+        Returns:
+            The pseudo-loss whose GGN is the MC-Fisher.
+        """
+        # Sample gradients w.r.t. output using the provided generator
+        grad_output_samples = sample_grad_output_flat(
+            output_flat.detach(), mc_samples, y, generator
+        )
```
runame: This function is identical to `c_pseudo_flat` for the EF, besides the sampling here (and the adjustment of the `reduction_factor` by multiplying with `mc_samples`). Might be worth it to avoid this code duplication if possible.

runame: Ah I missed this, thanks! I think I actually prefer efficiency here then. If you really want you could add a comment about this.

runame (approved):

PR #219, opened 2026-01-08
[OPT] Generalize eigenvalue correction trick to mild weight sharing

f-dangel:
This PR is a follow-up of #216  and generalizes its trick from `S=1` (no weight sharing) to `S=small`.
Runtimes of EKFAC remain the same in our benchmark. Peak memory is further reduced on ResNet18+CIFAR10.

### DIFF
--- a/docs/rtd/internals.rst
+++ b/docs/rtd/internals.rst
@@ -14,4 +14,4 @@ KFAC-related
 EKFAC-related
 -------------
 
-.. autofunction:: curvlinops.ekfac.compute_eigencorrection
+.. autofunction:: curvlinops.ekfac.compute_eigenvalue_correction_linear_weight_sharing


curvlinops/ekfac.py
```diff
@@ -27,6 +27,185 @@
 from curvlinops.kfac_utils import extract_patches
 
 
+def compute_eigencorrection(
```
runame: Nit: I prefer being overly explicit here and call it something like `compute_eigenvalue_correction`.

curvlinops/ekfac.py
```diff
@@ -27,6 +27,185 @@
...
+    g: Tensor, ggT_eigvecs: Tensor, a: Tensor, aaT_eigvecs: Union[Tensor, None]
+) -> Tensor:
+    r"""Computes eigen-corrections for a linear layer with weight sharing.
```
runame: Same here: "eigen-corrections" -> "eigenvalue corrections"

PR #208, opened 2025-12-20
Removed diag/trace estimators, modified example to use skerch, general CI fixes

andres-fr:
The main scope of this PR is to address the following issue:

**Scope and performance of diagonal approximations**  https://github.com/f-dangel/curvlinops/issues/198

Proposed changes in this PR
* Removed stochastic diagonal and trace estimators from the curvlinops codebase
* Removed subsequent tests and ensured API docs remain consistent
* Updated trace/diagonal example from the gallery to use skerch functionality instead
* Added `__rmatmul__` to the examples tensor linop to allow for skerch interoperability (shouldn't harm other linops from the core lib)
* Added `skerch` to the  test dependencies

As a result, trace and diagonal estimators are now totally out of the scope of curvlinops, but the gallery example shows that it integrates nicely with skerch, and the errors even got lower!

---

Additionally, this PR also proposes the following changes, which are not related to the issue above, and I needed in order to properly lint/test the repo and compile the docs:
* Raised python support from 3.9 to 3.10: Python 3.9 is deprecated and caused dependency issues
* Also fixed `pytest-optional-tests` version to 0.1.1, newer versions cause `make test` to crash
* Fixed a bunch of pydoctest errors, and excluded docs from pydoctest (many errors were IMO unfixable and wouldn't make sense to cover, plus pydoctest doesn't allow to exclude/include subdirectories as per e.g. [this](https://github.com/PyCQA/pydocstyle/issues/363) and [this](https://github.com/fannheyward/coc-pyright/issues/494) issues.
* Fixed CUDA code, which was broken in most examples (e.g. plotting CUDA tensors, which doesn't work)

As a result, all `make` commands now work from a fresh install on my end (Ubuntu system with CUDA GPU).

---

I managed to do all these changes fast enough to be on top of main HEAD, so although some of them are fundamental and out of scope, hopefully we can land on something that is good for the project. Happy to hear your feedback and adapt!

Also, happy holidays :)
Andres

PR #221, opened 2026-01-09
Removed implementation and test for trace and diagonal estimators, updated corresponding gallery example to use skerch

andres-fr:
The PR proposed in https://github.com/f-dangel/curvlinops/pull/208 had 3 main lines of contribution:

Updating Python to 3.10, since 3.9 is deprecated
General CI fixes
The actual content of the PR (removing diag/trace implementations and tests and using skerch in the gallery example)

It was agreed to split those PR into 3 subsequent, individual PRs.
The first and second one have already been merged to main.
The present PR is the last one of the bunch, i.e. removing the diag/trace implementations

Specifically:
* Removed trace and diag implementations  and autoimports from library
* Removed corresponding unit tests
* Modified gallery example to work with skerch
* Added skerch as doc dependency

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Main features:
 
 - **Broad support of curvature matrices.** `curvlinops` supports many common curvature matrices and approximations thereof, such as the Hessian, Fisher, generalized Gauss-Newton, and K-FAC ([overview](https://curvlinops.readthedocs.io/en/latest/linops.html#linear-operators), [visual tour](https://curvlinops.readthedocs.io/en/latest/basic_usage/example_visual_tour.html#visualization)).
 
-- **Unified interface.** All linear operators share the same interface, making it easy to switch between curvature matrices.
+- **Unified interface.** All linear operators share the same interface, making it easy to switch between curvature matrices. The linear operator interface also allows for seamless integration with other linop-based libraries, such as [`skerch`](https://github.com/andres-fr/skerch) ([example](https://curvlinops.readthedocs.io/en/latest/basic_usage/example_trace_diagonal_estimation.html)).
 
 - **Purely PyTorch.** All computations can run on a GPU.
 


docs/examples/basic_usage/example_trace_diagonal_estimation.py
```diff
@@ -280,189 +329,47 @@ def plot_estimation_results(
...
 
-    plt.savefig("diagonal_estimation.pdf", bbox_inches="tight")
+    plt.savefig("./diagonal_estimation.pdf", bbox_inches="tight")
+
 
 # %%
 #
-# For fast spectral decay, XDiag yields more accurate diagonal estimates than
-# vanilla Hutchinson. For slow spectral decay, its benefits disappear.
-#
```
f-dangel: Not true anymore?

andres-fr:
There is an issue of nondeterminism with the ekfac test.

In this PR the test failed when triggered by commit [Delete diagonal_estimation.pdf (now for good?)](https://github.com/f-dangel/curvlinops/pull/221/commits/d962732e7fcc99a51841fe32fd3a73e0ff411fbe)

But if you check the exact same commit in my forked repo (where the PR comes from), it actually passes [link to corresponding utest action](https://github.com/andres-fr/curvlinops/actions/runs/20866578778).

I reviewed curvlinops commit history, and there is nothing between both of them, so the code is identical and I assume the inconsistency must come from some nondeterminism.

To "investigate" this, I will run a dummy commit to trigger the test again.

f-dangel:
Yeah, I think you are right. It could be related to which Github runner is assigned to run the test action.

docs/examples/basic_usage/example_trace_diagonal_estimation.py
andres-fr: > Not true anymore?

I'll check the example so it makes sense.
But the statement is true regardless of implementation.

This is discussed in more detail in the Epperly PhD thesis, see e.g. section 14.7

https://arxiv.org/abs/2512.15929

docs/rtd/linops.rst
```diff
@@ -79,21 +79,6 @@ Spectral density approximation
...
-.. autofunction:: curvlinops.hutchpp_trace
-
-.. autofunction:: curvlinops.xtrace
-
-Diagonal approximation
-======================
-
-.. autofunction:: curvlinops.hutchinson_diag
-
-.. autofunction:: curvlinops.xdiag
```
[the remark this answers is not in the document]
andres-fr: With remove you mean like the other 2?
* Remove implementation
* Remove any associated utests
* Update any associated docs/ gallery examples with skerch alternative

And what do you mean with "frobenius norm approximation" exactly? the contents of `/curvlinops/norm`?

My understanding from analyzing the code is that it does not directly depend on the deleted diag/trace code (it does look like a Hutch reimplementation though).
If that is the case, I'd suggest to remove it in a separate PR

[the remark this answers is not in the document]
f-dangel: Sure, I can set up a separate PR after this one.

docs/examples/basic_usage/example_trace_diagonal_estimation.py
```diff
@@ -75,50 +114,43 @@ def create_power_law_matrix(dim: int = DIM, c: float = 1.0) -> Tensor:
         A sample matrix with a power law spectrum.
     """
     # Create the diagonal matrix Λ with Λii = i^(-c)
-    L = (arange(1, dim + 1, dtype=float64) ** (-c)).diag()
+    L = arange(1, dim + 1, dtype=dtype) ** (-c)
```
f-dangel: I am not a big fan of using broadcasting (the below `(Q.H * L) @ Q`) and would prefer if we kept `L` as a matrix, although this will slightly degrade performance.

docs/examples/basic_usage/example_trace_diagonal_estimation.py
```diff
@@ -29,38 +39,67 @@
...
+    assert num_meas % 3 == 0, "num_meas must be a multiple of 3"
+    # from xhutchpp we just use the deflation matrix Q here (num_meas // 3)
+    Q = _xhutchpp(lop, lop_device, lop_dtype, num_meas // 3, 0, seed - num_meas)["Q"]
+    result = _hutch(lop, lop_device, lop_dtype, num_meas // 3, seed, defl_Q=Q)
+    result["diag"] += (Q.T * (Q.H @ lop)).sum(0)  # here another num_meas // 3
+    result["tr"] = result["diag"].sum()
+    return result
+
+
+def xdiagtrace(lop, lop_device, lop_dtype, num_meas, seed):
```
f-dangel: Would be nice to document these methods, as they will likely be copy-pasted by users.

Issue #197, opened 2025-10-17
Implement adjoint matmul


andres-fr:
## Issue (feature/bug):

Given a linear operator `A`, Python provides the possibility of implementing  `__rmatmul__`  for adjoint multiplication (`v.H @ A`) as well as the more popular `__matmul__` (`A @ v`).

For Hermitian linops, such as most curvature linops, it is not *really* necessary to implement both, since `A = A.H` and therefore `v.H @ A = (A.H @ v).H = (A @ v).H`.

The issue is that **sometimes we may want to run a curvilnop on a routine that does not assume Hermitian symmetry**, and this is a legitimate practice (e.g. SVDs are generally known to be more stable than EIGH).

In those cases, the routine may feature adjoint multiplications, and it will not leverage Hermitian symmetry as shown above. And running those routines on curvlinops operators crashes. E.g. when running a sketched SVD with a curvlinops, the following line crashes:

```
# this line doesn't work! 
x = (Q.H @ lop)
# this works
x = lop @ Q
```
With the following error:

```
TypeError: unsupported operand type(s) for @: 'Tensor' and 'HessianLinearOperator'
```

## Fix:

Implement `__rmatmul__` for curvlinops. Based on the math above, the fix for Hermitian linops should be trivial, something like (untested):

```
def __rmatmul__(self, x):
    x = self.__matmul__(x)
    return x.H
```

PR #222, opened 2026-01-10
[REF] Merge adjoint and matmat/matvec tests

### LINKED ISSUE #183: [REF] Remove `adjoint` option in utilities for comparing `matmat`s
Both the functions `test.utils.compare_matmat` and `test.utils.compare_consecutive_matmats` currently have an option `adjoint` which, if enabled, takes the adjoints of the operators before comparing their matrix-matrix multiplication routines. This can lead to confusion (e.g. [here](https://github.com/f-dangel/curvlinops/pull/175#discussion_r1986167915)). Both functions should be simplified by removing this option. This requires extracting the logic for taking the adjoint into all tests.

f-dangel:
This PR merges tests that verify the correctness of matrix-vector multiplication and matrix-matrix multiplication for an operator and its transpose. This reduces the number of related tests by 4 (without affecting the test coverage), reducing the overall cost as we have to evaluate the ground truth curvature matrix less often. This PR therefore reduces the time to execute the tests, and also refactors the underlying functionality to resolve #183.

### DIFF
--- a/curvlinops/ggn.py
+++ b/curvlinops/ggn.py
@@ -11,8 +11,6 @@ from torch.nn import Module, Parameter
 from curvlinops._torch_base import CurvatureLinearOperator
 from curvlinops.utils import make_functional_model_and_loss
 
-s
-
 
 def make_ggn_vector_product(
     f: Callable[..., Tensor], c: Callable[..., Tensor], num_c_extra_args: int = 0


test/test__torch_base.py
```diff
@@ -200,41 +195,47 @@ def test_SumPyTorchLinearOperator(adjoint: bool, is_vec: bool):
     A_plus_B = A + B
     A_linop, B_linop = TensorLinearOperator(A), TensorLinearOperator(B)
     A_plus_B_linop = A_linop + B_linop
-    compare_matmat(A_plus_B_linop, A_plus_B, adjoint, is_vec)
+    compare_matmat(A_plus_B_linop, A_plus_B)
+
+    A_plus_B, A_plus_B_linop = A_plus_B.adjoint(), A_plus_B_linop.adjoint()
+    compare_matmat(A_plus_B_linop, A_plus_B)
```
f-dangel: @runame long-term, these explicit duplications that test multiplication with the adjoint operator will be absorbed back into `compare_matmat`. This will be possible once we implement #197, and then extend `compare_matmat` to test both `A @ x` and `x @ A` which relies on taking the adjoint as `x @ A = (A.adjoint() @ x.adjoint()).adjoint()`.

PR #226, opened 2026-01-27
[ADD] Support left multiplication (`__rmatmul__`) in linear operators

### LINKED ISSUE #197: Implement adjoint matmul

f-dangel:
Only review after #224.

This PR adds support for left-multiplication that supports both tensor and tensor list formats.
Resolves #197.

### DIFF
--- a/test/test__torch_base.py
+++ b/test/test__torch_base.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from typing import Iterable, Iterator, List, MutableMapping, Tuple, Union
 
 from pytest import raises
-from torch import Tensor, linspace, manual_seed, rand, rand_like, randperm, zeros
+from torch import Size, Tensor, linspace, manual_seed, rand, rand_like, randperm, zeros
 
 from curvlinops._torch_base import CurvatureLinearOperator, PyTorchLinearOperator
 from curvlinops.examples import TensorLinearOperator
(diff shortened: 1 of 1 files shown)


curvlinops/_torch_base.py
```diff
@@ -75,6 +75,47 @@ def __init__(
...
+            X, self._out_shape, free_dim
+        )
+
+        # Convert from leading to trailing dimension for compatibility with _matmat
+        # Pre-processing always adds explicit matrix dimension, so always use movedim
+        XH = [x.conj().movedim(0, -1) for x in X]
+
+        # Use that X @ A = (A^H @ X^H)^H
+        AH = self.adjoint()
+        AH_XH = AH @ XH
```
runame: what will the preprocessing in `__matmul__` do? if not both pre-/postprocessing steps were needed, we could directly call `self._matmat(XH)` here.

PR #231, opened 2026-02-04
[ADD] Generalize `hessian_sqrt` to sequence-valued predictions

f-dangel:
This PR generalizes the function that computes the Hessian square root of a loss w.r.t. the prediction to support sequence-valued predictions. It also adds dedicated unit tests for this function (both for sequence- and vector-valued predictions) as it was previously only indirectly tested through type-II KFAC.


runame (approved):
LGTM.

PR #232, opened 2026-02-06
[ADD] Make (E)KFAC's properties methods

f-dangel:
Makes (E)KFAC linear operators more similar in behaviour to PyTorch tensors (i.e. one can call `K.trace()` and get the trace of a tensor or linear operator). Removes caching their values.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support left multiplication with linear operators (`X @ A` with `X` a tensor or tensor list)
   ([PR](https://github.com/f-dangel/curvlinops/pull/226))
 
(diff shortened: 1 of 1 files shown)


curvlinops/ekfac.py
```diff
@@ -707,51 +700,43 @@ def det(self) -> Tensor:
...
+        return _det
 
-    @property
     def logdet(self) -> Tensor:
         r"""Log determinant of the EKFAC approximation.
 
-        More numerically stable than the ``det`` property.
+        More numerically stable than the ``det`` method.
         Will call ``compute_kronecker_factors`` and ``compute_eigenvalue_correction`` if
         either of them has not been called before and will cache the logdet until one of
```
runame: and here

f-dangel:
Thanks for catching! Should be fixed now.

PR #296, opened 2026-04-19
Support layer-wise damping for KFAC/EKFAC, which might be useful for adaptive damping and faster convergence.

liaodisen:
Inspired by the paper: https://arxiv.org/pdf/2112.02089, a regularized Newton Method, and can be used for KFAC/EKFAC
The idea is for each layer $i$, the damping is

$$
\lambda_i = \max(\lambda_{\min}, \sqrt{c \lVert g_i \rVert_2}),
$$

A simple use case:
```{python}
ADAPTIVE_DAMPING = GradientNormDamping(
    damping_scale=ADAPTIVE_SCALE, min_damping=MIN_DAMPING
)
logits = model(X)
loss = loss_function(logits, y)
loss.backward()

gradients = [p.grad.detach().clone() for p in params.values()]
damping = (
     ADAPTIVE_DAMPING(gradients) if use_adaptive_damping else FIXED_DAMPING
  )
```

**Results**
On Cifar10, we set $c = 0.1$ for adaptive damping, and $\lambda = 0.1$ for fixed damping, and train CNN and ResNet18 with identical settings for other hyperparameters:

adaptive_damping_convergence
adaptive_damping_small_cnn_convergence

### DIFF
--- a/curvlinops/ekfac.py
+++ b/curvlinops/ekfac.py
@@ -1,5 +1,7 @@
 """Contains LinearOperator implementation of EKFAC approximation of the Fisher/GGN."""
 
+from collections.abc import Sequence
+
 from curvlinops._torch_base import _ChainPyTorchLinearOperator
 from curvlinops.blockdiagonal import BlockDiagonalLinearOperator
 from curvlinops.computers._base import ParamGroup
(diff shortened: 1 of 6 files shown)


f-dangel:
Hi Disen, do we need to incorporate an official interface for this?

I think you can do this pretty easily in the current code, without needing an interface:
```python
# compute KFAC
KFAC = KFACLinearOperator(...)
# unpack, B is block-diagonal Kronecker-factored
P, B, PT = KFAC
# invert each block with its own damping
dampings = [...]
damped_inv_blocks = [b.inverse(d) for b, d in zip(B, dampings)]
# re-construct the block-diagonal op representing the per-layer-damped inverse
B_inv = BlockDiagonalLinearOperator(damped_inv_blocks)
# build inverse KFAC
KFAC_inv = P @ B_inv @ PT
```

Let me know if this code does what you were trying to achieve, and let me know should I have missed something.

f-dangel:
ping @liaodisen

liaodisen:
You are right, it is easy to achieve without changing the interface.
Maybe add the interface when this algorithm is provably better.

PR #297, opened 2026-04-20
support generic losses for TYPE2 Fisher.

liaodisen:
Modify the `ggn_utils.py` to support any loss function for  `loss_hessian_matrix_sqrt`


f-dangel:
Hi Disen, thanks for working on this! I had a quick look. Can we split up this PR so that we first generalize the core routines (one PR), before supporting general losses in linear operators like KFAC (second PR)?

liaodisen:
Will close this PR and open another one

